PR #76, opened 2023-11-09
Provide SCM information to changelog generator

fourls:
The keepachangelog Maven plugin expects SCM information to be provided in the top-level POM, and will not generate header link implementations if it can't find it. SonarDelphi's SCM information is erroneously in `sonar-delphi-plugin/pom.xml`, not `/pom.xml`, and so running `mvn keepachangelog:release` will remove all existing header links.

This PR moves the SCM information into the correct place and slightly adjusts the changelog to match the plugin's generated output.


Cirras (approved):
Looking good!

PR #77, opened 2023-11-09
Fix type resolution of constructor calls on types accessed through an array access

fourls:
When performing type resolution, SonarDelphi erroneously resolves the type of the `Create` call in this example to `TEnclosingType`:

```pascal
type
  TActualType = class(TObject)
  end;

  TEnclosingType = class(TObject)
    procedure MyProc;
  end;

procedure Foo(Obj: TActualType); overload;
begin
  WriteLn('Actual');
end;

procedure Foo(Obj: TEnclosingType); overload;
begin
  WriteLn('Enclosing');
end;

// ...

procedure TEnclosingType.Bar;
type
  TActualTypeClass = class of TActualType;
const
  CActualTypes: array of TActualTypeClass = (TActualType);
begin
  Foo(CActualTypes[0].Create); // Should resolve to TActualType, but resolves to TEnclosingType instead
end;
```

This bug discards any elements in a primary expression that are prior to a series of name references. and was identified from a number of false positive NonLinearCast issues. This PR fixes this bug and adds some extra tests to NonLinearCast to ensure these cases are caught.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   expression.
 - Improve name resolution for declarations within types.
 - Improve type resolution for array accesses into variants.
+- Improve type resolution around constructor calls following array accesses in primary expressions.
 - Improve parsing and type modeling around `AnsiString` types with specified code pages.
 
 ### Removed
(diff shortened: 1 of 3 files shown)


Cirras (approved):
Nice detective work in finding this. 👍

Cirras (approved):

PR #83, opened 2023-11-10
Skip integration tests by default

Cirras:
Because integration tests spin up a SonarQube server, they share SonarQube's Java requirements.

SonarQube only runs under Java 17 (no other versions are supported), meaning that the integration tests require the developer to be running JDK 17 *exactly*.

This is a weird/inconvenient limitation to impose on contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ read the [contributing guide](docs/CONTRIBUTING.md).
 
 ## Development
 
-SonarDelphi can be built with JDK 17+ using [Maven](https://maven.apache.org/).
+SonarDelphi can be built with JDK 11+ using [Maven](https://maven.apache.org/).
 
 To build the project and run unit tests, execute the following command from the project's root directory:
 
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ read the [contributing guide](docs/CONTRIBUTING.md).
 
 SonarDelphi can be built with JDK 17+ using [Maven](https://maven.apache.org/).
 
-To build the project and run all tests, execute the following command from the project's root directory:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -115,6 +115,18 @@ mvn clean install
 
 After building, the plugin jar can be found in `sonar-delphi-plugin/target`.
 
+### Integration Tests
```
fourls, line 118: I think this might be a little too prominent, particularly with the callout. A single sentence like "If you are using JDK 17, integration tests can be run by passing `-DskipTests=false` as a parameter." or similar might get the same meaning across in a more concise way.

The other option is to not say anything - I'm on the fence about whether it's important enough to be in the README at all.

README.md
```diff
@@ -115,6 +115,18 @@ mvn clean install
...
+
+By default integration tests are skipped. To run them, use `-DskipITs=false`:
```
fourls, line 120: We do say "all tests" in the previous section. Maybe reword that to say

> To build the project and run all **unit** tests, execute the following command from the project's root directory:

pom.xml
fourls, line 1: Using the CI profile still skips the integration tests, including in [GH action run](https://github.com/integrated-application-development/sonar-delphi/actions/runs/6820399627/job/18549232053). 

image

It works as expected passing `-DskipITs=false`. Is it possible that the child pom overrides the properties from the parent pom, even properties specified in profiles, while passing as a command line arg overrides both?

fourls (changes requested):
The CI profile still skips ITs, but it looks like a good approach overall - just a few things about wording.

fourls (approved):
Looks good!

PR #87, opened 2023-11-16
Exclude class reference types from inline declaration search

### LINKED ISSUE #86: [Bug]: Importing a type that contains a class reference type referencing itself causes a stack overflow
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

_No response_

### Issue description

Delphi allows you to declare class reference types *inside* the class that is being referenced. For example:

```delphi
type
  TFoo = class(TObject)
  type
    TFooClass = class of TFoo;
  end;
```
If this unit is imported SonarDelphi reads the interface section, then checks for inline type declarations to see if the implementation section needs to be read as well. This step causes a stack overflow, as the type scope of `TFooClass` is the same as `TFoo`.

Stack trace:

image

### Steps to reproduce

I have provided a minimal case project below. (Ignore the inaccurate name, should be SelfClassReferenceCrashTestProject). 

[SelfTypeAliasCrashTestProject.zip](https://github.com/integrated-application-development/sonar-delphi/files/13371006/SelfTypeAliasCrashTestProject.zip)

fourls:
Resolves #86 using the discussed approach, in which the scopes of class reference types are not traversed when finding inline declarations.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/SymbolTableBuilder.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/SymbolTableBuilder.java
@@ -42,6 +42,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


Cirras (approved):
👍

PR #89, opened 2023-11-17
Upload plugin jar artifact in build workflow

fourls:
This PR adds a step to the build workflow to save the generated sonar-delphi-plugin jar as an artifact. This saves having to check out the repository and run a `mvn install` when retrieving a specific version of the jar.


Cirras (approved):
Good idea 👍

PR #91, opened 2023-11-19
Update `VisibilityKeywordIndentation` to refer to 'leading' whitespace

zaneduffield:
The rule description erroneously refered to 'trailing' whitespace, but it checks indentation which is actually leading whitespace.


fourls (approved):
Totally right!

PR #93, opened 2023-11-23
Sequential sections

### LINKED ISSUE #65: New rule: Sequential sections of the same type should be merged
### Prerequisites

### Suggested rule title

Sequential sections of the same type should be merged

### Rule description

It would detect sequential blocks/sections of the same type, including

- `var` blocks
- `const` blocks
- visibility sections

For example:

```delphi
type
  TFoo = class(TObject)
  private
    FName: string;
  private // Noncompliant
    FAge: Integer;
    procedure Bar;
  public
    procedure Baz;
  end;
```

```delphi
type
  TFoo = class(TObject)
  private // Compliant
    FName: string;
    FAge: Integer;
    procedure Bar;
  public
    procedure Baz;
  end;
```

### Rationale

When two identical blocks are non-sequential, there's a solid chance that there could be a reason - e.g. defining a constant, variable, or property that is used in later sections - but when they are sequential, having the two sections separate is pointless and harms readability.

If the behaviour of this rule is a little overloaded, it could also be split into different sub-rules for different types of section?

fourls:
Resolves #65 by adding four new rules:

- ConsecutiveTypeSection, which creates issues for consecutive `type` sections
- ConsecutiveVarSection, which creates issues for consecutive `var` and consecutive `threadvar` sections
- ConsecutiveConstSection, which creates issues for consecutive `const` sections and consecutive `resourcestring` sections
- ConsecutiveVisibilitySection, which creates issues for consecutive visibility sections of the same level, excluding sections with only one item that is a `type`, `const`, or `class var` (the rule does also catch consecutive cases of these)

This PR also makes a couple of API improvements to facilitate these rules, particularly:

* The type of a `const` section (`const` or `resourcestring`) is now exposed
* The type of a `var` section (`var` or `threadvar`) is now exposed


delphi-checks/src/main/java/au/com/integradev/delphi/checks/ConsecutiveVisibilitySectionCheck.java
```diff
@@ -0,0 +1,158 @@
...
+      return Optional.empty();
+    }
+  }
+
+  @Override
+  protected FilePosition getViolatingFilePosition(DelphiNode node) {
+    if (node instanceof VisibilitySectionNode) {
+      DelphiNode firstChild = node.getChild(0);
+      if (firstChild instanceof VisibilityNode) {
+        return FilePosition.from(firstChild);
```
Cirras, line 116: We can safely do these casts and also safely assume that the `VisibilitySectionNode` has a visibility node, because:
* We control what gets passed to this method in `areViolatingSections`
* We select the 2nd section as the violating node, and an `IMPLICIT_PUBLISHED` visibility section can only ever be the first one.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/ConsecutiveConstSection.html
```diff
@@ -0,0 +1,23 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Having consecutive <code>const</code> sections is useless and negatively impacts readability.
+  Combining the sections into one larger section is easier to understand and more economical.
```
Cirras, line 4: Let's change the wording on "more economical" in these descriptions.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/ConsecutiveVisibilitySection.html
```diff
@@ -0,0 +1,41 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Having consecutive visibility sections with the same visibility is useless and negatively
+  impacts readability.
+  Combining the sections into one larger section is easier to understand and more economical.
+</p>
+<p>
```
Cirras, line 7: Should this have a `<h3>Exceptions</h3>` header?

Cirras (changes requested):
Really good, just a few comments to follow up. 👍

Cirras (approved):

fourls:
Oh, did we want these in the Sonar way profile? @Cirras

Cirras:
> Oh, did we want these in the Sonar way profile? @Cirras

Yes, can you open a new PR?

PR #118, opened 2023-12-04
Fix support for NUnit 2.0 XML schema when parsing test results

fourls:
SonarDelphi is currently reading NUnit test reports using the NUnit 3.0 spec, but DUnitX's exported reports use NUnit 2.0.
The main difference is that the test result strings are different between the two versions - e.g. a passed NUnit 2.0 test uses "Success" while NUnit 3.0 uses "Passed". This was causing DUnitX tests to erroneously be reported as "Failed" in the SonarQube UI. 

This PR adds support for NUnit 2.0 test reports *in addition to* NUnit 3.0. It will automatically detect whether a given NUnit report is v2 or v3 and parse accordingly.


Cirras (changes requested):

Cirras (approved):
👍

PR #119, opened 2023-12-04
Use semantic analysis in CastAndFree

### LINKED ISSUE #115: CastAndFree rule misidentifies most method calls as casts
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

_No response_

### Issue description

The CastAndFree rule is intended to catch cases where a cast is performed immediately prior to calling `Free` or `FreeAndNil`. Currently, it also picks up almost any routine call with an argument list as a cast and incorrectly flags them as redundant casts. This commonly occurs in code where the intention is to immediately discard the return value of a function.

This is due to this problematic code in `CastAndFreeCheck.java`:

```java
private static boolean isHardCast(ExpressionNode expr) {
  return expr instanceof PrimaryExpressionNode
      && expr.getChild(0) instanceof NameReferenceNode
      && expr.getChild(1) instanceof ArgumentListNode
      && expr.getChildren().size() < 6;
  }
```

I believe this rule predates semantic analysis, hence its inaccuracy - it could be easily rewritten to be accurate using SonarDelphi's modern semantic analysis tools (like in `AbstractCast` and descendants).

### Steps to reproduce

Scan a Delphi file including the following:

```delphi
function DoSomething(Obj: TObject): TObject;
begin
  Result := Obj;
end;

procedure Test;
var
 MyObj: TObject;
begin
  FreeAndNil(DoSomething(TObject.Create));
end;
```

_No response_

fourls:
Fixes #115 by using type and symbol table information to evaluate casts in the CastAndFree rule.

Currently, our cast logic is implemented entirely in the AbstractCast base rule, and is only available to descendants of said class. This PR extracts the cast detection logic from AbstractCast into a new utility, CastUtils, which parses primary expressions and binary expressions into cast objects. CastAndFree was then reimplemented using CastUtils.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Stack overflow on class reference types that reference their containing type.
 - Scan failures on redundant unit aliases in .dproj files.
 - Incorrect file position calculation for multiline compiler directives.
+- Incorrect detection of method calls as hard casts in `CastAndFree`.
 
 ## [1.0.0] - 2023-11-14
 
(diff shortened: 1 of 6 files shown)


Cirras (approved):
Great! 👍

PR #121, opened 2023-12-05
Exclude routines, properties, and fields with attributes from dead code checks

### LINKED ISSUE #113: Test methods marked with an attribute are reported as unused
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

_No response_

### Issue description

Many test frameworks (such as DUnitX) use attributes to mark test case methods. These are incorrectly marked by SonarDelphi as unused. 

Because elements marked with an attribute are significantly more likely to be accessed via RTTI, we probably shouldn't be assuming that just because they have no direct references they are not used at all. The most correct approach would be to explicitly all ignore routines, properties, etc. with an attached attribute, since we can't statically verify that they are unused.

### Steps to reproduce

Scan the following code:

```delphi

uses DUnitX.TestFramework;

type
  [TestFixture]
  TMyTest = class(TObject)
  public
    [TestCase]
    procedure TestMyCase; // Incorrectly marked as unused - it is executed and run at runtime
  end;

initialization
  TDUnitX.RegisterTestFixture(TMyTest);
```

_No response_

fourls:
Routines, properties, or fields that are annotated with an attribute are generally done so because they will be accessed at runtime via RTTI. This PR fixes #113 by excluding these members from UnusedRoutine, UnusedProperty, and UnusedField.

A couple of changes have been made to the API to facilitate this change:

- `AttributeListNode` now has a `getAttributeTypes` method that returns the list of attributes as their types
- `PropertyNameDeclaration` now has a `getAttributeTypes` method that returns a list of the types of each annotated attribute
- `RoutineNameDeclaration` now has a `getAttributeTypes` method that returns a list of the types of each annotated attribute

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ConstSectionNode::isResourceStringSection` API method.
 - `AttributeListNode::getAttributeTypes` API method.
 - `RoutineNameDeclaration::getAttributeTypes` API method.
+- `PropertyNameDeclaration::getAttributeTypes` API method.
 
 ### Changed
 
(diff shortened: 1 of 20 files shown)


Cirras (approved):
🎉

Issue #81, opened 2023-11-10
Handle NativeInt as a weak alias in Delphi 12


fourls:
### Prerequisites

- [X] This improvement has not already been suggested.
- [X] This improvement would be generally useful, not specific to my code or setup.

### Engine area

Delphi language support

### Improvement description

In Delphi 11 and below, NativeInt is a "strong alias" that is equivalent to either Integer or Int64 (depending on the platform). In Delphi 12, the NativeInt data type is a "weak alias" that is **exactly equal to** Integer (on Win32) or Int64 (on Win64).

Because strong aliases have less weight when performing overload resolution, this change needs to be reflected in SonarDelphi. 
This should be accompanied with retooled support for type aliases, as there is currently no way to retrieve a type alias instead of the aliased type, and without full type alias support, SonarDelphi would not be able to detect NativeInt usages at all. This would have ramifications on several rules including PlatformDependentTruncation and PlatformDependentCast.

For more details, see [Marco Cantu's blog post describing the change](https://blogs.embarcadero.com/yukon-beta-blog-nativeint-as-a-weak-alias/) and [the RAD Studio documentation for System.NativeInt](https://docwiki.embarcadero.com/Libraries/Athens/en/System.NativeInt).

### Rationale

This is required to have accurate overload resolution on Delphi 12 code.

Cirras:
The difficult part of this is now complete. 👍
Weak aliases are modeled in the type system as of b3b14dd58f27e1b5b5f5b65dfb53f37aced1b012.

zaneduffield:
Note: the symmetric change was _not_ made to the `NativeUInt` type in Delphi 12

zaneduffield:
Actually, I tested that theory using the wrong version of Delphi 🤦. `NativeUInt` was actually changed to a weak alias type as well (despite the documentation suggesting it was only `NativeInt` that changed).

PR #128, opened 2023-12-10
Improve type resolution on expressions involving integers

Cirras:
This PR significantly improves type resolution for binary and unary expressions with integer operands.

Changes are based on the following data tables:
- [binary](https://github.com/integrated-application-development/sonar-delphi/issues/107#issuecomment-1841868071)
- [unary](https://github.com/integrated-application-development/sonar-delphi/issues/107#issuecomment-1840094351)

Other notable changes:
- Only intrinsic operator overloads that are **relevant** to the operands will be considered
  - (e.g. adding 2 records should not consider the intrinsic `(Integer, Integer): Integer` overload)
- Subranges of integers are now modeled more correctly as `IntegerType` instances
- Corrected the name of the bitwise `not` operator from `BitwiseNot` to `OnesComplement`


delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
```diff
@@ -270,58 +320,129 @@ private Set<Invocable> createAnd() {
...
 
-    return addAll(
-        createIntegerArithmeticBinary(name),
-        new OperatorIntrinsic(name, List.of(extended, extended), extended),
-        new OperatorIntrinsic(name, List.of(integer, extended), extended),
-        new OperatorIntrinsic(name, List.of(extended, integer), extended));
+    Type nativeInt = typeFactory.getIntrinsic(IntrinsicType.NATIVEINT);
+    Type nativeUInt = typeFactory.getIntrinsic(IntrinsicType.NATIVEUINT);
+
+    if (!nativeInt.isWeakAlias()) {
```
zaneduffield, line 389: This could be `nativeInt.isStrongAlias()`. Did you choose to negate the `weak` version because that was the terminology used in the Delphi 12 release?

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
```diff
@@ -270,58 +320,129 @@ private Set<Invocable> createAnd() {
     Type uint32 = typeFactory.getIntrinsic(IntrinsicType.CARDINAL);
     Type uint64 = typeFactory.getIntrinsic(IntrinsicType.UINT64);
 
+    Type widenedUInt16 = ((TypeFactoryImpl) typeFactory).widenedUInt16();
+    Type widenedUInt32 = ((TypeFactoryImpl) typeFactory).widenedUInt32();
```
zaneduffield, line 324: I don't understand the names here.
It doesn't make sense as a UInt32 that has been widened, because it's a smaller range than that of UInt32.

zaneduffield (changes requested):
Looking really good overall. I'm impressed how easy you made it look to perfectly emulate the operator overload behaviour.

My only real feedback is regarding some naming.

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
Cirras: It's actually because SonarDelphi doesn't model the unique `Native*` types as strong aliases at all - but rather as distinct integer types with the same value range as the platform-specific integer types.

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
Cirras: Agreed, these names are a little awkward.

The idea was to express:
- "unsigned integer that has been widened to a 16-bit value"
and
- "unsigned integer that has been widened to a 32-bit value"

I don't love the current names, but also don't have many alternate ideas.

Thoughts?

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
zaneduffield: Right, I was thinking of `isStrongAlias` and `isWeakAlias` and being the negative of one another because I was just looking at the impl in `AliasTypeImpl`. The types could of course not be aliases at all.

I guess this is dead code until we do #81.

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
zaneduffield: The host type for the `0..High(Integer)` type is `Integer`, so it feels wrong to put `UInt` in the name of these.
Even if it is _semantically_ better to think of these as unsigned integers, the width given doesn't make sense; the `widenedUInt32` type doesn't have 32 usable bits, it has 31.

I'm also not sure what the best names for these are, but I feel like the current names are misleading in more than one way:
* what exactly is being 'widened' is not clear
* the width in the name doesn't correspond to the width of the actual type

What do you think about `anonymousUInt15` and `anonymousUInt31`?

Cirras: Thanks! I went with `anonymousUInt*`

zaneduffield (approved):
👍

PR #130, opened 2023-12-11
Add support Unicode identifiers

### LINKED ISSUE #125: Support Unicode identifiers
### Prerequisites

- [X] This improvement would be generally useful, not specific to my code or setup.

### Engine area

Delphi language support

## Analysis

In Delphi, some Unicode characters may be used in identifiers.

Currently we only support having surrogate pairs in places where Unicode identifier characters should be allowed.
https://github.com/integrated-application-development/sonar-delphi/blob/51a3969f306a38e7d63376417c4af5206ba73e95/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g#L1228-L1230

The [documentation](https://docwiki.embarcadero.com/RADStudio/en/Fundamental_Syntactic_Elements_(Delphi)#Identifiers) has this to say on the topic
> An identifier must begin with an alphabetic character, a Unicode character, or an underscore (_) and cannot contain spaces. Alphanumeric characters, Unicode characters, digits, and underscores are allowed after the first character.

It leaves 'a Unicode character' up to interpretation, but from reverse-engineering I've found that an identifier can start with any codepoint in the ranges:
1. [U+41, U+5A] ('A'-'Z')
2. [U+5F, U+5F] ('_')
3. [U+61, U+7A] ('a'-'z')
4. [U+80, U+2FFF]
5. [U+3001, U+D7FF]
6. [U+E000, U+FF0F]
7. [U+FF1A, U+FFFF]

To summarise: any ASCII alphabetic character, underscore, or codepoint from the [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) (i.e. no surrogate pairs) that isn't an ideographic space (U+3000, considered whitespace) or a full-width numeral (U+FF10 - U+FF19). 

An identifier can also be continued with an ASCII numeral (U+30 - U+39) or a full-width numeral (U+FF10 - U+FF19).

The strange rules around the ideographic characters is currently undocumented, but I recently raised [RSP-42431](https://quality.embarcadero.com/browse/RSP-42431) to have this fixed.

## Suggested Implementation

We should definitely include _most_ characters from the [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) in the grammar for `Alpha`, but how many of the special cases we account for is up for debate.

I think that we should match the compiler in not permitting the full-width numerals in identifiers, since there's potential for lexical ambiguity with integer literals (despite the fact that the compiler doesn't let you use them in integer literals). On the other hand I think we should permit surrogate pairs in identifiers, because there's nothing else that they could be.

### Case-Insensitive Comparison

Including a wider range of Unicode characters has implications on case-insensitive comparison. As always, the details of this as the compiler understands it are not fully documented. From my investigation, it appears that the compiler does a Unicode-aware case-insensitive comparison (that may or may not be dependent on locale and Unicode normalisation).

I'm not saying that we currently handle this incorrectly. It's just something to check while we're fixing this.

<hr/>

Should fix #124.

### Rationale

The full range of valid Delphi identifiers should be supported.

zaneduffield:
Fixes #125

The Unicode-aware case-insensitive identifier comparison that we are currently performing doesn't _perfectly_ match the Delphi compiler, but @Cirras and I decided it wasn't worth trying to match this unspecified behaviour. Instead, I documented the situation with some tests and comments in `DelphiScopeImplTest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for unicode characters in identifiers.
 - Support for the NUnit 2.0 test results format (in addition to existing 3.0 support).
 - `ConsecutiveVarSection` analysis rule, which flags consecutive `var` sections that can be merged.
 - `ConsecutiveConstSection` analysis rule, which flags consecutive `const` sections that can be
--- a/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
+++ b/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
@@ -1189,7 +1189,7 @@ TkArrayIndices          : 'ARRAY_INDICES'
 //****************************
 // Tokens
 //****************************
-TkIdentifier            : (Alpha | '_') (Alpha | Digit | '_')*
+TkIdentifier            : (Alpha | '_') (Alpha | FullWidthNumeral | Digit | '_')*
                         ;
                         // We use a lookahead here to avoid lexer failures on range operations like '1..2'
                         // or record helper invocations on Integer literals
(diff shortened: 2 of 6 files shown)


Cirras (dismissed):
Great work, thanks for tackling this one. 🎉

Cirras (approved):

PR #135, opened 2023-12-12
Highlight the whole hard cast expression when reporting issues

Cirras:
The current standard behavior around flagging hard casts is to highlight the argument list in the issue.
```
ShortInt(123);
        ^^^^^
```

This can be very confusing in the case of nested hard casts, and you might erroneously think the issue applies to the inner cast.
 Consistently highlighting the whole cast expression would make this clearer:
```
ShortInt(123);
^^^^^^^^^^^^^
```

This PR changes the behavior to highlight the whole hard cast expression in the following rules:
- `UnicodeToAnsiCast`
- `CharacterToCharacterPointerCastCheck`
- `NonLinearCastCheck`
- `RedundantCastCheck`
- `PlatformDependentCast`

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PlatformDependentCastCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PlatformDependentCastCheck.java
@@ -21,12 +21,14 @@ package au.com.integradev.delphi.checks;
 import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.plugins.communitydelphi.api.ast.ArgumentListNode;
+import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.NameReferenceNode;
 import org.sonar.plugins.communitydelphi.api.ast.Node;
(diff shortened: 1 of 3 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/PlatformDependentCastCheck.java
```diff
@@ -54,7 +56,18 @@ public DelphiCheckContext visit(ArgumentListNode argumentList, DelphiCheckContex
         Type castedType = getHardCastedType(argumentList, context);
```
fourls, line 56: Oops, I must have forgotten to migrate PlatformDependentCast to use CastUtils. Would it be possible for you to do this as part of this change? I can also do it in a separate PR if preferred.

fourls (commented):
Looks good, just the one note about CastUtils.

I'm not particularly keen on the commit message, it sounds a little clunky to me. Maybe `Highlight entire hard cast expression in cast checks`?

fourls:
I've also tested this change and can confirm that it works as expected 👍

delphi-checks/src/main/java/au/com/integradev/delphi/checks/PlatformDependentCastCheck.java
Cirras: Separate PR preferred on this one.

fourls (approved):
Looks good!

fourls (approved):

PR #139, opened 2023-12-14
Provide default value of `excludeApi` properties to SonarQube

Cirras:
A minor oversight in #137 - we didn't provide a default value to SonarQube for the new `excludeApi` properties.

This PR provides a default value of "false" for all `excludeApi` rule properties, making that default value explicit in the SonarQube UI and also making it clearer that the property is a boolean value.


fourls (approved):
I don't love the implicit boolean->string conversion, but there's not much you can do about it when it's a constant expression. 

Looks good to me!

Issue #108, opened 2023-11-25
Bogus report of UnusedLocalVariable


sglienke:
### Prerequisites

- [X] This bug is in SonarDelphi, not SonarQube or my Delphi code.
- [X] This bug has not already been reported.

### SonarDelphi version

1.0.0

### SonarQube version

10.3.0.82913 Community Edition

### Issue description

Following local routine in Spring.pas (around line 6010) raises an `UnusedLocalVariable` on `rightValue`:

```
  function RawEquals(const recordType: TRttiType): Boolean;
  var
    leftRec, rightRec: Pointer;
    field: TRttiField;
    leftValue, rightValue: TValue;
  begin
    if left.TypeInfo = right.TypeInfo then
    begin
      if IsManaged(left.TypeInfo) then
      begin
        leftRec := left.GetReferenceToRawData;
        rightRec := right.GetReferenceToRawData;
        for field in recordType.GetFields do
        begin
          leftValue := field.GetValue(leftRec);
          rightValue := field.GetValue(rightRec);
          if not leftValue.Equals(rightValue) then
            Exit(False);
```

### Steps to reproduce

Running SonarDelphi on Spring4D 2.0.0

sglienke:
Possibly the same cause as #104 - this would also explain another occurrence.
In the following code `elType` is marked as unused:

```
function ConvDynArray2DynArray(const source: TValue; target: PTypeInfo; var value: TValue;
  const formatSettings: TFormatSettings): Boolean; //FI:O804
var
  len, i: Integer;
  res, v1, v2: TValue;
  elType: PTypeInfo;
begin
  len := source.GetArrayLength;
  MakeDynArray(target, len, res);

  elType := target.TypeData.DynArrElType^;
```

Cirras:
Agreed that this is most likely related to (or the same as) #104.

PR #140, opened 2023-12-14
Fix name resolution failures around type helpers for weak aliases

### LINKED ISSUE #104: Variable being passed as untyped out parameter is not noticed as being initialized
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

10.3.0.82913 Community Edition

### Issue description

SonarDelphi raises `community-delphi:VariableInitialization` on the following code (Spring.ValueConverters.pas line 1540 - from 2.0.0 tag)

```
function TDateToStringConverter.DoConvertTo(const value: TValue;
  const targetTypeInfo: PTypeInfo; const parameter: TValue): TValue;
var
  format: TFormatSettings;
begin
  if not parameter.IsEmpty and parameter.TryAsType(TypeInfo(TFormatSettings), format) then
  begin
    case targetTypeInfo.Kind of
      tkString, tkUString:
        Result := TValue.From<string>(DateToStr(value.AsExtended, format)); // <-
```

The logic about this code is that via `parameter` a `TFormatSettings` will be passed and TryAsType has this signature:

`function TValueHelper.TryAsType(typeInfo: PTypeInfo; out target): Boolean;` - when it returns true then the out parameter is being filled. Since the type is being passed via the PTypeInfo parameter the out needs to be typeless.

### Steps to reproduce

Running SonarDelphi on Spring4D 2.0.0

_No response_

Cirras:
Internally, we map `Type` -> `HelperType` when doing helper lookups in a given scope.

The analyzer models weak alias types by generating a transparent wrapper around the aliased type. It looks and behaves exactly like the aliased type, but the two type objects are not **equal** to each other.

The source of truth about a type's identity is the image, so the solution here is to simply map `<type image>` -> `HelperType` instead.

Fixes #104 and #108.

### DIFF
--- a/delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/helpers/WeakAliasHelper.pas
+++ b/delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/helpers/WeakAliasHelper.pas
@@ -1,0 +1,34 @@
+unit WeakAliasHelper;
+
+interface
+
+type
+  TFoo = record
+    procedure Bar;
+  end;
+
+  TFooAlias = TFoo;
+
+  TFooHelper = record helper for TFooAlias
+    procedure Bar;
+  end;
+
+implementation
+
+procedure TFoo.Bar;
+begin
+  // Do nothing
+end;
+
+
+procedure TFooHelper.Bar;
+begin
+  // Do nothing
+end;
+
+procedure Test(Foo: TFoo);
+begin
+  Foo.Bar;
+end;
+
+end.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Scan failures on redundant unit aliases in .dproj files.
 - Incorrect file position calculation for multiline compiler directives.
 - Incorrect detection of method calls as hard casts in `CastAndFree`.
+- Name resolution failures around helpers extending weak alias types.
 
 ## [1.0.0] - 2023-11-14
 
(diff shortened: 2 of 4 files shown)


delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/helpers/WeakAliasHelper.pas
```diff
@@ -0,0 +1,32 @@
+unit Simple;
+
+interface
+
+type
+  TFoo = record
+    procedure Bar;
+  end;
+
+  TFooHelper = record helper for TFoo
```
fourls, line 10: I think you forgot to make this a weak alias 😄 I checked and this test already passes on master.

Updating this file to match the below does fail on master and pass in this branch, though, so the main code change is all good.
```delphi
type
  TFoo = record
    procedure Bar;
  end;

  TFooAlias = TFoo;

  TFooHelper = record helper for TFooAlias
    procedure Bar;
  end;
```

delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/helpers/WeakAliasHelper.pas
```diff
@@ -0,0 +1,32 @@
+unit Simple;
```
fourls, line 1: This unit header should be named `WeakAliasHelper` to match the filename.

fourls (commented):
Main code change looks good! Only thing is that the test does not cover the change.

delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/helpers/WeakAliasHelper.pas
Cirras: Whoops, I'm a dork.
Fixing that...

fourls (approved):
All good now! I'll merge once the checks pass.

PR #143, opened 2023-12-14
Fix name resolution failures when `Inc`/`Dec` are invoked on pointers

### LINKED ISSUE #141: Name resolution failures when `Inc` and `Dec` are invoked on pointers
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

10.3

### Issue description

`Inc` and `Dec` can be used on typed pointers, but the analyer currently only handles the case where the first argument is an ordinal type.

```java
routine("Dec").varParam(ANY_ORDINAL).param(type(INTEGER)).required(1).returns(type(INTEGER));
routine("Inc").varParam(ANY_ORDINAL).param(type(INTEGER)).required(1);
```

An additional overload is needed for `Inc` and `Dec` that accepts typed pointers.

### Steps to reproduce

See code example.

```pas
procedure Test(P: PByte);
var
  I: Integer; // UnusedLocalVariable issue will be raised here
begin
  I := 5;
  Inc(P, I);
end;
```

Cirras:
This PR adds a new `TYPED_POINTER` overload to the `Inc` and `Dec` intrinsic signatures.

Fixes #141.

### DIFF
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/UnusedLocalVariableCheckTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/UnusedLocalVariableCheckTest.java
@@ -21,6 +21,8 @@ package au.com.integradev.delphi.checks;
 import au.com.integradev.delphi.builders.DelphiTestUnitBuilder;
 import au.com.integradev.delphi.checks.verifier.CheckVerifier;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class UnusedLocalVariableCheckTest {
   @Test
@@ -208,6 +210,26 @@ class UnusedLocalVariableCheckTest {
                 .appendImpl("procedure Test(Foo: Integer);")
                 .appendImpl("begin")
                 .appendImpl("  // Do nothing")
+                .appendImpl("end;"))
+        .verifyNoIssues();
+  }
+
+  // See: https://github.com/integrated-application-development/sonar-delphi/issues/141
+  @ParameterizedTest
+  @ValueSource(strings = {"Inc", "Dec"})
+  void testIssue141ShouldNotAddIssue(String routine) {
+    CheckVerifier.newVerifier()
+        .withCheck(new UnusedLocalVariableCheck())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendDecl("type")
+                .appendDecl("  PByte = ^Byte;")
+                .appendImpl("procedure Test(P: PByte);")
+                .appendImpl("var")
+                .appendImpl("  I: Integer;")
+                .appendImpl("begin")
+                .appendImpl("  I := 5;")
+                .appendImpl(String.format("  %s(P, I);", routine))
                 .appendImpl("end;"))
         .verifyNoIssues();
   }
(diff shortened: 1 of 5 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcher.java
```diff
@@ -41,6 +41,11 @@ public final class IntrinsicArgumentMatcher extends TypeImpl {
           "<ordinal>",
           type -> type.isInteger() || type.isBoolean() || type.isEnum() || type.isChar());
 
+  public static final Type ANY_TYPED_POINTER =
+      new IntrinsicArgumentMatcher(
+          "<typed pointer>",
+          type -> type.isPointer() && ((PointerType) type).dereferencedType().size() != 0);
```
fourls, line 47: Is there a reason you're using `dereferencedType().size() != 0` instead of `dereferencedType().isUntyped()`?

delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcherTest.java
```diff
@@ -83,6 +84,14 @@ void testAnyOrdinal() {
     assertThat(matches(ANY_ORDINAL, FACTORY.getIntrinsic(IntrinsicType.STRING))).isFalse();
   }
 
+  @Test
+  void testAnyTypedPointer() {
+    Type typedPointer = FACTORY.pointerTo("PInteger", FACTORY.getIntrinsic(IntrinsicType.INTEGER));
+    assertThat(matches(ANY_TYPED_POINTER, typedPointer)).isTrue();
+    assertThat(matches(ANY_TYPED_POINTER, FACTORY.nilPointer())).isFalse();
+    assertThat(matches(ANY_TYPED_POINTER, FACTORY.nilPointer())).isFalse();
```
fourls, line 92: Duplicate line - did you mean to test `FACTORY.untypedPointer()` in the second case here?

fourls (commented):
Just a few little comments.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcher.java
fourls: I see this would fail in the nil pointer case - in that case, maybe the following?
It is unintuitive to me that a size of 0 means "untyped or nil" - especially since size doesn't really make sense in these scenarios. I still think it would be good to have it be more explicit.

```java
type -> type.isPointer() && !((PointerType) type).dereferencedType().isUntyped() && !((PointerType) type).isNilPointer());
```

delphi-checks/src/test/java/au/com/integradev/delphi/checks/UnusedLocalVariableCheckTest.java
```diff
@@ -211,4 +213,24 @@ void testUnusedArgumentShouldNotAddIssue() {
                 .appendImpl("end;"))
         .verifyNoIssues();
   }
+
+  // See: https://github.com/integrated-application-development/sonar-delphi/issues/141
```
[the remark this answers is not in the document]
Cirras: This one is a regression test, and really only exists to ensure that #141 is not reintroduced in the future.

fourls (approved):
Looks good!

PR #144, opened 2023-12-18
Clarify wording of VariableName description

### LINKED ISSUE #96: Improve description of `VariableName` rule
### Prerequisites

### Rule to improve

VariableName

### Improvement description

In the `VariableName` rule description, further clarify that the rule applies to *all variables* and visually separate the note about global variables requiring a `G` prefix.

### Rationale

We received the following feedback on [ADUG](https://forums.adug.org.au/t/sonardelphi-v1-0-0-released/60111/5):

>  I got like 22k issues of which most looked like parsing errors and mistakenly taking parameterless const or var parameters as global variables telling me I should stick to the naming convention of prefixing them with a G

These issues were all raised because the variables were lowercase and didn't comply with PascalCase.
The user's confusion here indicates that the rule description isn't clear enough.

fourls:
Fixes #96 by rephrasing the VariableName description, emphasising that the rule enforces naming conventions for both local and global variables.


Cirras (approved):
Great!

PR #147, opened 2023-12-20
Improve behavior of various intrinsic routines

Cirras:
This PR:
- Improves type comparisons between text types
  - `AnsiString` -> `UnicodeString` conversions are preferred over
    `UnicodeString` -> `AnsiString` conversions.
  - `UnicodeString` -> `Variant` conversions are preferred over
    `UnicodeString` -> `AnsiString` conversions.
  - `AnsiString` -> `UnicodeString` conversions are preferred over
    `AnsiString` -> `Variant` conversions.
  - ⭐ discovered while fixing the `Concat` intrinsic, which was originally going to have `(AnsiString, AnsiString)` and `(UnicodeString, UnicodeString)` overloads to disambiguate
- Improves type conversions from character pointers to strings
  - `PAnsiChar` -> `AnsiString` conversions are preferred over `PAnsiChar` -> `ShortString`.
  - ⭐ important for the `Length` intrinsic, which has `AnsiString` and `ShortString` overloads
- Fixes various intrinsic signatures
  - incorrect return types (`Concat`, `Copy`,  `Slice`, `Succ`)
  - incorrect parameter types (`Assert`, `Assign`, `AssignFile`, `Sqr`)
  - missing overloads (`Copy`, `Length`)
  - overly-strict `var` parameter typings (`var` parameters are expected to have an exact type-match with the argument)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Return type of the array `Copy` intrinsic was not specified correctly.
 - Dynamic array literals were not accepted as arguments to `Copy`.
 - Strong aliases to `System.Text` were not accepted as arguments to `Append`.
+- Implicit conversions to `Boolean` were not accepted as arguments to `Assert`.
 
 ## [1.0.0] - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Strong aliases were not accepted as arguments to `VarArrayRedim`, `VarCast`, `VarClear`, or
   `VarCopy`.
 - Parameter of the `Int64` overload of `Sqr` was erroneously specified as `Real`.
+- Return type of the `Succ` intrinsic was not specified correctly.
 
 ## [1.0.0] - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dynamic array literals were not accepted as arguments to `Concat`.
 - Return type of the array `Copy` intrinsic was not specified correctly.
 - Dynamic array literals were not accepted as arguments to `Copy`.
+- Strong aliases to `System.Text` were not accepted as arguments to `Append`.
 
 ## [1.0.0] - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Strong aliases were not accepted as arguments to `Delete`, `Insert`, or `SetLength`.
 - Strong aliases and typed pointers were not accepted as arguments to `Dispose`, `FreeMem`,
   `GetMem`, `New`, or `ReallocMem`.
+- Strong aliases and specialized string types were not accepted as arguments to `Str`, `GetDir`,
+  `SetString`, or `Val`.
 
 ## [1.0.0] - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve type resolution on binary and unary expressions.
 - Improve type comparisons between signed and unsigned integer types.
+- Improve type comparisons between text types.
 - Exclude routines annotated with attributes in `UnusedRoutine`.
 - Exclude properties annotated with attributes in `UnusedProperty`.
 - Exclude fields annotated with attributes in `UnusedField`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dynamic array literals were not accepted as arguments to `Copy`.
 - Strong aliases to `System.Text` were not accepted as arguments to `Append`.
 - Implicit conversions to `Boolean` were not accepted as arguments to `Assert`.
+- Strong aliases and typed files were not accepted as arguments to `Assign`.
 
 ## [1.0.0] - 2023-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,9 +57,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dynamic array literals were not accepted as arguments to `Concat`.
 - Return type of the array `Copy` intrinsic was not specified correctly.
 - Dynamic array literals were not accepted as arguments to `Copy`.
-- Strong aliases to `System.Text` were not accepted as arguments to `Append`.
 - Implicit conversions to `Boolean` were not accepted as arguments to `Assert`.
-- Strong aliases and typed files were not accepted as arguments to `Assign`.
+- Strong aliases were not accepted as arguments to `Append`, `SeekEof`, `SeekEoln`, or `SetTextBuf`.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Available for `UnusedConstant`, `UnusedField`, `UnusedGlobalVariable`, `UnusedProperty`,
     `UnusedRoutine`, and `UnusedType`.
   - Excludes public API (declared with public visibility in the interface section).
+- `ShortString` and `AnsiString` overloads for `Length` intrinsic.
```
fourls, line 25: I think this should be in the "Fixed" section. Very technically they're additions, but in practice it's a fix for something we were modelling incorrectly (just like the other things in this PR that have been put under "Fixed").

fourls: Didn't mean to complete the review so fast 🤦 I'm still reviewing.

Cirras: Makes sense to me. 👍

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcher.java
```diff
@@ -20,12 +20,29 @@
...
+import org.sonar.plugins.communitydelphi.api.type.IntrinsicType;
 import org.sonar.plugins.communitydelphi.api.type.StructKind;
 import org.sonar.plugins.communitydelphi.api.type.Type;
 
 public final class IntrinsicArgumentMatcher extends TypeImpl {
-  public static final Type ANY_DYNAMIC_ARRAY =
-      new IntrinsicArgumentMatcher("<dynamic array>", Type::isDynamicArray);
+  public static final Type LIKE_DYNAMIC_ARRAY =
+      new IntrinsicArgumentMatcher(
+          "<like dynamic array>",
```
fourls, line 30: Might be good to name this `<dynamic array, array constructor, string, or char>`, just because
`assertion failed: expected <like dynamic array>` is a bit of a confusing message.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcher.java
```diff
@@ -20,12 +20,29 @@
...
+          type ->
+              type.isDynamicArray()
+                  || type.isArrayConstructor()
+                  || type.isString()
+                  || type.isChar());
+
+  public static final Type ANY_STRING =
+      new IntrinsicArgumentMatcher("<string>", type -> type.isString() || type.isChar());
```
fourls, line 38: In the same vein, this should be `<string or char>`.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java
```diff
@@ -118,4 +142,99 @@ public Type getReturnType(List<Type> arguments) {
...
+      if (result == null) {
+        result = typeFactory.arrayConstructor(elementTypes);
+      }
+
+      return result;
+    }
+
+    private Type widenIfNeeded(Type newType, @Nullable Type currentType) {
+      if (newType.isChar()) {
+        newType = typeFactory.getIntrinsic(newType.size() == 1 ? ANSISTRING : UNICODESTRING);
```
fourls, line 175: What's the context of using `newType.size() == 1` here? Looks to me like this is only true for AnsiChar?

delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcherTest.java
```diff
@@ -18,55 +18,126 @@
...
+    assertThat(matches(ANY_FILE, FACTORY.fileOf(FACTORY.getIntrinsic(IntrinsicType.INTEGER))))
+        .isTrue();
+    assertThat(matches(ANY_FILE, FACTORY.getIntrinsic(IntrinsicType.INTEGER))).isFalse();
+  }
+
+  @Test
+  void testAnyTextFile() {
+    assertThat(matches(ANY_TEXT_FILE, FACTORY.untypedFile())).isFalse();
+    assertThat(matches(ANY_TEXT_FILE, FACTORY.getIntrinsic(IntrinsicType.TEXT))).isTrue();
+    assertThat(matches(ANY_TEXT_FILE, FACTORY.fileOf(FACTORY.getIntrinsic(IntrinsicType.INTEGER))))
```
fourls, line 103: I assume we've confirmed that e.g. `file of string` doesn't get counted as a text file by Delphi?

delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcherTest.java
```diff
@@ -18,55 +18,126 @@
...
+    Type fixedArray = FACTORY.array(null, elementType, Set.of(ArrayOption.FIXED));
+    Type dynamicArray = FACTORY.array(null, elementType, Set.of(ArrayOption.DYNAMIC));
+    Type openArray = FACTORY.array(null, elementType, Set.of(ArrayOption.OPEN));
+    Type arrayOfConst =
+        FACTORY.array(null, elementType, Set.of(ArrayOption.OPEN, ArrayOption.ARRAY_OF_CONST));
+
+    assertThat(matches(ANY_ARRAY, fixedArray)).isTrue();
+    assertThat(matches(ANY_ARRAY, dynamicArray)).isTrue();
+    assertThat(matches(ANY_ARRAY, openArray)).isTrue();
+    assertThat(matches(ANY_ARRAY, arrayOfConst)).isTrue();
```
fourls, line 129: Maybe test that `ANY_ARRAY` doesn't match strings.

CHANGELOG.md
```diff
@@ -32,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve type resolution on binary and unary expressions.
 - Improve type comparisons between signed and unsigned integer types.
+- Improve type comparisons between text types.
+- Improve type conversions from character pointers to strings.
```
fourls, line 37: This doesn't sound quite right to me. Maybe "Improve accuracy of" instead of just "Improve"?

fourls (commented):
Whew, that's a lot! Looks pretty good overall, a lot of QoL and correctness fixes that I'm sure will have run-on effects in reducing name resolution problems.

I can't verify that much of the logic is correct, particularly around type comparisons, but I have cross-referenced the intrinsic routines that have been changed with the docs and all seems to be in order.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java
Cirras: > this is only true for AnsiChar?

We're just abusing this detail to avoid an unnecessary string comparison.
I don't mind changing it for clarity, though.

Thoughts?

delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicArgumentMatcherTest.java
Cirras: `file of string` is not allowed, actually.
`E2155 Type 'string' needs finalization - not allowed in file type`

`Text` is really a sort of weird aberration, it's different from the other file types in a few ways.
It doesn't even have the same size as a normal `file of X` type.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java
fourls: I think it would be good to change it for clarity. This could also invisibly be broken at some point in the future, since the code's intention differs from its implementation.

fourls (approved):
Looks good!

PR #148, opened 2023-12-21
Make file encoding detection consistent with the `sonar-scanner` engine

### LINKED ISSUE #136: File encoding mismatch between SonarQube and sonar-delphi APIs
### Prerequisites

### SonarDelphi version

1.0.0

### SonarQube version

_No response_

### Issue description

Currently, sonar-delphi simply uses the filesystem-level encoding provided by the sonar-scanner engine.
However, the sonar-scanner engine actually determines file encodings at the file level, using heuristics:

https://github.com/SonarSource/sonarqube/blob/f50873318397d4bf7ba6a5c2b194dfa02492bdae/sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/MetadataGenerator.java#L53
https://github.com/SonarSource/sonarqube/blob/f50873318397d4bf7ba6a5c2b194dfa02492bdae/sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/ByteCharsetDetector.java#L48

This encoding is then saved on the `InputFile` which can be accessed via [`InputFile::charset`](https://javadocs.sonarsource.org/6.5/apidocs/org/sonar/api/batch/fs/InputFile.html#charset--)

This can lead to [errors](https://github.com/integrated-application-development/sonar-delphi/issues/124#issuecomment-1846895757) in file offsets, when sonar-delphi is decoding the file using a different charset than the sonar-scanner engine.

### Steps to reproduce

Scan the following (utf-8 encoded) source file
```delphi
program LineOffsetBug;

begin
                           // LEcranPrincipal doit être la 1ère vraie form créée pour être la main form fenêtre principale
end.
```

with `sonar.sourceEncoding` set to windows-1252. 

_No response_

Cirras:
This PR fixes #136.
- If a BOM is detected, the corresponding encoding will **always** be used.
- Otherwise:
  - If we're constructing the `DelphiFile` based on an `InputFile`...
    - we use [InputFile::charset](https://javadocs.sonarsource.org/6.5/apidocs/org/sonar/api/batch/fs/InputFile.html#charset--)
  - If we **aren't** constructing the `DelphiFile` based on an `InputFile`...
    - we fall back to `sonar.sourceEncoding`
    - if it isn't provided, we fall back to `Charset::defaultCharset`


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- File encoding detection was inconsistent with the `sonar-scanner` engine.
```
fourls, line 48: From a user's perspective, I don't think the problem was necessarily that the detection was inconsistent, but that it was blindly applying the encoding rather than using a heuristic. Maybe something like `Source encoding configuration was overriding the encoding detection heuristics in all cases.`?

fourls (commented):
Apart from a changelog nitpick, all looks good to me! I might do a little testing and confirm that this now causes SonarDelphi to work correctly in the situation that caused #136, but after that I'm happy to merge.

fourls (approved):

fourls:
Works a charm!

PR #154, opened 2024-01-05
Tidy up intrinsic return type inference

Cirras:
This PR ties up some loose ends around intrinsics.

There are still a few issues around intrinsic return type inference:
* `Copy` wasn't returning string types for character pointer or variant arguments
* `Concat` wasn't widening character types to string types (for single-character string literals)

Also, I added tests for the `IntrinsicReturnType` meta-types.


fourls (approved):
Looks good. Good thinking adding the return type tests.

PR #155, opened 2024-01-07
Migrate from `json-simple` to `gson`

Cirras:
`json-simple` is unmaintained and has an annoying API that encourages unsafe casts.
It also pulls in a vulnerable version of junit 4 as a compile dependency, generating meaningless noise from Sonatype Lift.

(**note:** there is no actual vulnerability, junit wasn't even supposed to be a compile dependency in `json-simple`)

### DIFF
--- a/delphi-checks/pom.xml
+++ b/delphi-checks/pom.xml
@@ -60,8 +60,8 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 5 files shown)


fourls (approved):
Looks good, I tested loading the plugin on SonarQube, adding a custom rules plugin, and running a scan - all worked well with no errors.

PR #158, opened 2024-01-11
Downgrade duplicate unit import error to warning

fourls:
Delphi `requires` clauses permit duplicate units, e.g. a `.dpk` with the following code compiles:

```delphi
package MyPackage;

requires
  rtl,
  vcl,
  rtl;

end.
```

SonarDelphi does not currently permit this. This PR adds a special case to `DelphiScopeImpl::addDeclaration` that catches the duplicate declaration exception and logs a warning instead if the declaration is a unit import.

Note that this change also "permits" duplicate imports in normal uses clauses, but since this is a compilation error anyway I think it's the most practical way to solve this problem. Happy to hear other thoughts.


Cirras (approved):

Cirras:
Going to go ahead and merge, ci is only failing due to upstream issues in the SonarSource artifactory.

Cirras (approved):

PR #159, opened 2024-01-11
Support library units with no `begin`

fourls:
Delphi DLL units do not need a `begin`, but SonarDelphi's grammar currently requires them to have one. This means that the following valid `.dpr` fails to parse:

```delphi
library MyLib;

uses System.SysUtils;

end.
```

The `exports` section can also exist in DLL units without a `begin`, such as the following valid `.dpr`:

```delphi
library MyLib;

uses System.SysUtils;
exports XYZ;

end.
```

[JCL](https://github.com/project-jedi/jcl/blob/939d275863ba936df7fd7e9f65f41c829e46756d/jcl/packages/d29/JclRepositoryExpertDLL.dpr#L57) is a notable example of a project with DLL units in this format.

This PR updates the grammar to support this, and adds a test to ensure that these files can be parsed.


fourls:
CI failures unrelated... I'll re-run them later.

fourls:
After looking into this a little more, it seems like the `begin` can also be omitted in `program` files. 

I've updated the grammar to permit this for both `program` and `library` files - I have also confirmed that `package` files must not have a `begin`.

Cirras (approved):

Cirras (approved):

PR #166, opened 2024-01-24
Add checks to detect forms/frames without accompanying resource files

Cirras:
**New rules:**
* `FormDfm`: “VCL forms and frames should have a .dfm form file”
* `FormFmx`: “FireMonkey forms and frames should have a .fmx form file”

Neither rule is included in `Sonar way`.

**Compiler directives API:**
* The `RESOURCE` directive is modeled with a new `ResourceDirective` type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
+- **API:** `CompilerDirectiveParser` now returns a new `ResourceDirective` type when parsing
+  [resource directives](https://docwiki.embarcadero.com/RADStudio/en/Resource_file_(Delphi)).
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
 - `FormDfm` analysis rule, which flags VCL forms/frames that lack a `.dfm` resource.
+- `FormFmx` analysis rule, which flags FireMonkey forms/frames that lack a `.fmx` resource.
 - **API:** `CompilerDirectiveParser` now returns a new `ResourceDirective` type when parsing
   [resource directives](https://docwiki.embarcadero.com/RADStudio/en/Resource_file_(Delphi)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
+- `FormDfm` analysis rule, which flags VCL forms/frames that lack a `.dfm` resource.
 - **API:** `CompilerDirectiveParser` now returns a new `ResourceDirective` type when parsing
   [resource directives](https://docwiki.embarcadero.com/RADStudio/en/Resource_file_(Delphi)).
 
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/AbstractFormResourceCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/AbstractFormResourceCheck.java
@@ -1,0 +1,81 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2024 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.checks;
+
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+import org.sonar.plugins.communitydelphi.api.ast.DelphiAst;
+import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
+import org.sonar.plugins.communitydelphi.api.ast.TypeDeclarationNode;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.directive.CompilerDirective;
+import org.sonar.plugins.communitydelphi.api.directive.ResourceDirective;
+import org.sonar.plugins.communitydelphi.api.token.DelphiToken;
+import org.sonar.plugins.communitydelphi.api.type.Type;
+
+abstract class AbstractFormResourceCheck extends DelphiCheck {
+  protected abstract String getFrameworkName();
+
+  protected abstract String getFormTypeImage();
+
+  protected abstract String getFrameTypeImage();
+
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
+- `FormDfm` analysis rule, which flags VCL forms/frames lacking a `.dfm` resource.
+- `FormFmx` analysis rule, which flags FireMonkey forms/frames lacking a `.fmx` resource.
+- **API:** `ResourceDirective` type, which can be returned by the `CompilerDirectiveParser`.
```
fourls, line 16: This is very specific, but not very descriptive - I don't think anyone reading the changelog
will understand what this means in a concrete sense. Maybe something like

```markdown
- **API: `CompilerDirectiveParser` now returns a new `ResourceDirective` type when parsing resource directives (`{$R}`)
```

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
+- `FormDfm` analysis rule, which flags VCL forms/frames lacking a `.dfm` resource.
+- `FormFmx` analysis rule, which flags FireMonkey forms/frames lacking a `.fmx` resource.
```
fourls, line 15: Just a minor wording thing, maybe `that lack` rather than `lacking`?

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FormDfm.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "VCL forms and frames should have a .dfm form file",
```
fourls, line 2: Perhaps "should correspond to" instead of "should have"?

fourls: Or "should have a corresponding"?

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormDfmCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+class FormDfmCheckTest {
+  @Test
+  void testNormalClassWithoutDfmShouldNotAddIssue() {
+    CheckVerifier.newVerifier()
+        .withCheck(new FormDfmCheck())
+        .withSearchPathUnit(createVclForms())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendDecl("type")
+                .appendDecl("  TFoo = class(TForm)")
```
fourls, line 34: For clarity, might be a good idea to change the parent class `TForm` here to something else,
e.g. `TObject`. (This test only passes because it can't resolve the type.)

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormDfmCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+import au.com.integradev.delphi.builders.DelphiTestUnitBuilder;
+import au.com.integradev.delphi.checks.verifier.CheckVerifier;
+import org.junit.jupiter.api.Test;
+
+class FormDfmCheckTest {
+  @Test
+  void testNormalClassWithoutDfmShouldNotAddIssue() {
+    CheckVerifier.newVerifier()
+        .withCheck(new FormDfmCheck())
+        .withSearchPathUnit(createVclForms())
```
fourls, line 30: No need to add Vcl.Forms to the search path in this test.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormDfmCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendDecl("uses")
+                .appendDecl("  Vcl.Forms;")
+                .appendDecl("type")
+                .appendDecl("  TFoo = class(TForm) // Noncompliant")
+                .appendDecl("  end;"))
+        .verifyIssues();
```
fourls, line 83: I think it would be good to have some testing to confirm that:

- form in a file with other non-resource directives adds an issue
- form in a file with a non-DFM resource directive adds an issue
- etc. 

That's the most complex logic in the check and it's not really being tested here.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormFmxCheckTest.java
```diff
@@ -0,0 +1,110 @@
...
+                .appendDecl("type")
+                .appendDecl("  TFoo = class(TFrame)")
+                .appendDecl("  end;")
+                .appendImpl("{$R Foo.fmx}"))
+        .verifyNoIssues();
+  }
+
+  @Test
+  void testFormWithoutDfmShouldAddIssue() {
+    CheckVerifier.newVerifier()
```
fourls, line 73: As mentioned in the DFM check test, I think it would be good to have some testing to confirm that:

- form in a file with other non-resource directives adds an issue
- form in a file with a non-FMX resource directive adds an issue
- etc. 

That's the most complex logic in the check and it's not really being tested here.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FormFmxCheck.java
```diff
@@ -0,0 +1,71 @@
...
+import org.sonar.plugins.communitydelphi.api.ast.TypeDeclarationNode;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.directive.CompilerDirective;
+import org.sonar.plugins.communitydelphi.api.directive.ResourceDirective;
+import org.sonar.plugins.communitydelphi.api.token.DelphiToken;
+import org.sonar.plugins.communitydelphi.api.type.Type;
+
+@Rule(key = "FormFmx")
+public class FormFmxCheck extends DelphiCheck {
```
fourls, line 35: There's a really significant overlap between these two checks - could it be refactored
as two children of a base check class? I worry SonarCloud will complain about the level
of duplication here.

fourls (commented):
Looks good for the most part! Apart from a couple of testing gaps, I'm mostly just concerned that we'll get pinged for high duplication.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormDfmCheckTest.java
Cirras: Whoops, I didn't mean to inherit from an unresolved type here. Good catch.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FormFmxCheck.java
Cirras: I feel like a base class for such simple checks might be overkill.
The core logic *could* be extracted out though, that is true.

I should set up [branch analysis](https://docs.sonarsource.com/sonarcloud/enriching/branch-analysis-setup/) so we can get some certainty about these metrics instead of speculating.

fourls:
Ah, good catch!

CHANGELOG.md
Cirras: Ended up going for:
- **API: `CompilerDirectiveParser` now returns a new `ResourceDirective` type when parsing 
  [resource directives](https://docwiki.embarcadero.com/RADStudio/en/Resource_file_(Delphi)).

Cirras:
Added `ResourceDirective::getPredicates`.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/AbstractFormResourceCheck.java
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+
+    return context;
+  }
+
+  private boolean isFormResource(CompilerDirective directive) {
+    return directive instanceof ResourceDirective
+        && StringUtils.containsIgnoreCase(
+            ((ResourceDirective) directive).getResourceFile(), "." + getResourceFileExtension());
```
fourls, line 73: This could get triggered by a resource with the extension somewhere in it, e.g.
a file with only `Resources.FMXUtils.rc` would erroneously not trigger the FMX check.
`StringUtils.endsWithIgnoreCase` might be the way to go here.

fourls (commented):
Looks good! I think introducing the base class was a good move. Only noticed one thing.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/AbstractFormResourceCheck.java
Cirras: Makes sense to me. This would be a good test case as well.

Cirras: Added test cases to both rules for this.

fourls (approved):
Looks good!

PR #167, opened 2024-01-24
Allow unit imports to be shadowed

### LINKED ISSUE #161: Scan errors when defining a type with the same name of an imported unit
### Prerequisites

### SonarDelphi version

1.1.0

### SonarQube version

9.9.3

## Problem
When defining a `type` called the same as an imported library, `sonar-delphi` scan crashes.
This happens even if `sonar.delphi.unitScopeNames` is correctly populated.

## Expected
A rule catching this code smell
Related to #62

### Steps to reproduce

- Configure `sonar.delphi.unitScopeNames` to `WinApi`
- Execute a scan on this code snippet:
```pascal
uses
  Windows, // comment me when you uncomment "Winapi.Windows"
  //Winapi.Windows, -> uncomment me and the analysis will not crash
  System.SysUtils;

type
  TSample = class(TObject)
  private
  end;

  Windows = TSample;

begin
  try
    Writeln('Hello World!');
  except
    on E: Exception do
      Writeln(E.ClassName, ': ', E.Message);
  end;
end.
```

- The execution will raise the exception:
  `au.com.integradev.delphi.symbol.scope.DuplicatedDeclarationException: Windows is already in the symbol table`

[SonarQube_test.zip](https://github.com/integrated-application-development/sonar-delphi/files/13917240/SonarQube_test.zip)

Inside the zip there is a `sonar.log` file containing all the execution output in debug mode

fourls:
In Delphi, symbols can be declared with the same name as an imported unit. The behaviour seems to be that unit imports are shadowed by any subsequent declaration, but shadow all previous declarations. This PR implements this behaviour in SonarDelphi, which currently fails if a file-level symbol is declared with the same name as a unit import.

Fixes #161

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
@@ -605,6 +605,7 @@ public class NameResolver {
 
     disambiguateIsCallable();
     disambiguateVisibility();
+    disambiguateUnitImport();
 
     if (!isLastName) {
       disambiguateImplicitEmptyArgumentList();
@@ -1108,6 +1109,28 @@ public class NameResolver {
 
   private void disambiguateVisibility() {
     declarations.removeIf(not(this::isVisibleDeclaration));
+  }
+
+  private void disambiguateUnitImport() {
+    Set<UnitImportNameDeclaration> unitImports =
+        declarations.stream()
+            .filter(UnitImportNameDeclaration.class::isInstance)
+            .map(UnitImportNameDeclaration.class::cast)
+            .collect(Collectors.toSet());
+
+    if (unitImports.isEmpty()) {
+      return;
+    }
+
+    if (unitImports.stream().anyMatch(NameDeclaration::isImplementationDeclaration)) {
+      // The implementation unit import is chosen over anything in the interface
+      declarations.removeIf(not(NameDeclaration::isImplementationDeclaration));
+    }
+
+    if (unitImports.size() < declarations.size()) {
+      // The unit import is shadowed by anything below it
+      declarations.removeAll(unitImports);
+    }
   }
 
   private void disambiguateNotGeneric() {
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
@@ -170,6 +170,14 @@ public class DelphiScopeImpl implements DelphiScope {
 
     Set<NameDeclaration> duplicates = declarationsByName.get(declaration.getImage());
 
+    // Unit imports can clash with other declarations, except other imports
+    if (declaration instanceof UnitImportNameDeclaration) {
+      return duplicates.stream().noneMatch(UnitImportNameDeclaration.class::isInstance);
+    } else {
+      // Disregard unit imports when checking for duplicates
+      duplicates.removeIf(UnitImportNameDeclaration.class::isInstance);
+    }
+
     if (declaration instanceof GenerifiableDeclaration) {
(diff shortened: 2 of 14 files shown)


fourls:
Confirmed this passes our previously failing internal QA check.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
```diff
@@ -170,6 +170,14 @@ private boolean isAcceptableDuplicate(NameDeclaration declaration) {
 
     Set<NameDeclaration> duplicates = declarationsByName.get(declaration.getImage());
 
+    // Unit imports can clash with other declarations, except other imports
+    if (declaration instanceof UnitImportNameDeclaration) {
+      return duplicates.stream().noneMatch(UnitImportNameDeclaration.class::isInstance);
+    } else {
+      // Disregard unit imports when checking for duplicates
+      duplicates.removeIf(UnitImportNameDeclaration.class::isInstance);
```
Cirras, line 178: `duplicates` is a mutable set returned by `TreeMultimap::get`.
Modifying it in any way will modify the multimap, so we should do this differently.

Perfect-world, in a way that doesn't involve making a copy of the set.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
```diff
@@ -170,6 +170,14 @@ private boolean isAcceptableDuplicate(NameDeclaration declaration) {
 
     Set<NameDeclaration> duplicates = declarationsByName.get(declaration.getImage());
 
+    // Unit imports can clash with other declarations, except other imports
+    if (declaration instanceof UnitImportNameDeclaration) {
+      return duplicates.stream().noneMatch(UnitImportNameDeclaration.class::isInstance);
```
Cirras, line 175: I may be misreading your comment - but shouldn't this be `allMatch` if duplicate `UnitImportNameDeclaration`s are allowed?

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
```diff
@@ -1110,6 +1111,28 @@ private void disambiguateVisibility() {
     declarations.removeIf(not(this::isVisibleDeclaration));
   }
 
+  private void disambiguateUnitImport() {
+    Set<UnitImportNameDeclaration> unitImports =
+        declarations.stream()
+            .filter(UnitImportNameDeclaration.class::isInstance)
+            .map(UnitImportNameDeclaration.class::cast)
+            .collect(Collectors.toSet());
```
Cirras, line 1119: Prefer `toUnmodifiableSet`, since `unitImports` is queried but not modified.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
fourls: Ah, my comment is ambiguously worded. I mean to say "Unit imports can coexist with other declarations, except other imports". The code is correct.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
fourls: I've refactored to store and modify the stream instead.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/scope/DelphiScopeImpl.java
fourls: Reworded.

Cirras (approved):

PR #168, opened 2024-01-25
Support alternative casings for `WriteLn` and `ReadLn`

### LINKED ISSUE #102: Add alternative casings for `WriteLn` and `ReadLn` intrinsics
### Prerequisites

### SonarQube version

_No response_

### Rule to improve

MixedNames

### Improvement description

Recognize `Writeln` and `Readln` as alternative casings for `WriteLn` and `ReadLn`.

### Rationale

From debug info, we can clearly see that the `WriteLn` and `ReadLn` intrinsics are cased with an uppercase L internally.

```
 0001:00004958       System.@WriteLn
 0001:00004328       System.@ReadLn
```

However, code completion in the IDE refers to them as `Writeln` and `Readln`, and the documentation for [WriteLn](https://docwiki.embarcadero.com/Libraries/en/System.Writeln) and [ReadLn](https://docwiki.embarcadero.com/Libraries/en/System.Readln) use the same casing.

There are arguably 2 canonical casings for these intrinsics, so `MixedNames` should recognize them both.

```pas
Writeln('hello world'); // MixedNames issue will be raised on this line
Readln; // MixedNames issue will be raised on this line
```

fourls:
As described in #102, the jury is out on the casing of the `WriteLn` and `ReadLn` intrinsics - the debug information says `Ln`, while the documentation and the IDE say `ln`. For the purposes of the MixedNames check, SonarDelphi currently only accepts `WriteLn` and `ReadLn`.

This PR amends the MixedNames check to also accept the `ln` casing, and adds an additional (only partially documented) overload to the `ReadLn` intrinsic that does not include a file parameter. This was causing SonarDelphi to fail to resolve `Readln`and accidentally correctly accept it in some cases.

Fixes #102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for the `LLVM` symbol, which is defined on LLVM-based toolchains from Delphi 12 onward.
 - Support for the `IOSSIMULATOR` symbol, which is defined on the `DCCIOSSIMARM64` toolchain.
 
+### Changed
+
+- Alternative casings `Writeln` and `Readln` are now allowed in `MixedNames`.
+
 ### Fixed
 
 - Parsing errors on `.dpr` files without a top-level `begin`.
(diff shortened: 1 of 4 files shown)


Cirras:
> adds an additional (only partially documented) overload to the ReadLn intrinsic that does not include a file parameter

Nice one. Looks like this overload just reads standard input.

Cirras:
2 things:
* Added changelog entries, what do you think?
* Thoughts on the commit message "Add standard input overload for ReadLn intrinsic" for the first commit?

fourls:
Sounds good to me! I've updated it with that commit message now, so should be ready to merge.

Cirras (approved):
👍

PR #171, opened 2024-01-25
Add rules to validate format strings

### LINKED ISSUE #94: New rule: Format parameters should match the format string
### Prerequisites

### Suggested rule title

Format parameters should match the format string

### Rule description

This rule would catch calls to `System.SysUtils.Format` that have a mismatch between the specifiers in the format string and the provided arguments.

For example:

```delphi
Format('My name is %s and I am %d years old', ['Bob']); // Noncompliant - wrong argument count
Format('My name is %s', [5]); // Noncompliant - wrong type
```

Because this would involve parsing the format string, it may be relatively involved. With this in mind it could even potentially be two rules:

- Format arguments should match the number of format specifiers
- Format arguments should match the type of the corresponding format specifier

### Rationale

Passing invalid arguments to `Format` is very easy to do due to the lack of compile time checking, but raises a runtime exception.
It can lead to unexpected bugs - for example, `Format('I am %s years old', [MyAge])` looks fine but will raise an exception if called.

fourls:
As mentioned in #94, format strings are a common cause of unexpected runtime exceptions in Delphi. This PR adds a utility to parse format strings, as well as three rules using this feature:

* `FormatParameterCount`, which raises an issue if there is a mismatch between the expected and actual number of arguments to `Format`
* `FormatParameterType`, which raises an issue if there is a mismatch between the expected and actual types of arguments to `Format`
* `FormatStringValid`, which raises an issue if there is a syntax error in the format string

When used together, these rules should theoretically catch any case in which a format string is invalid (provided the format string is a single string literal - all these rules exclude cases in which the format string is complex).

Fixes #94


delphi-checks/src/main/java/au/com/integradev/delphi/utils/FormatStringUtils.java
```diff
@@ -0,0 +1,429 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2023 Integrated Application Development
```
Cirras, line 3: 2024

delphi-checks/src/main/java/au/com/integradev/delphi/utils/FormatStringUtils.java
```diff
@@ -0,0 +1,429 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public final class FormatStringUtils {
```
Cirras, line 33: Let's think about the API a bit more.

I don't love that we've got a bunch of imports that look like:
```java
import au.com.integradev.delphi.utils.FormatStringUtils.DelphiFormatArgument;
import au.com.integradev.delphi.utils.FormatStringUtils.DelphiFormatString;
import au.com.integradev.delphi.utils.FormatStringUtils.FormatSpecifierType;
```
Specifically, I'm not a big fan of `Utils` classes that export types and expect you to drag them out and use them directly. Grouping classes together is what packages are for, and we can just as easily hide the private classes with package-private visibility.

I wonder if it would make more sense to:
- pull all of these nested types out into a `utils.format` package
- make `FormatStringParser::parse` return a straight-up `DelphiFormatString`

I think that's the minimum shape we'd need this in, if we ever wanted to expose format string parsing via the public API in the future.

Cirras (changes requested):
Great work!
A few comments here and there, most of the things I noticed were quite minor.
Most important bit of feedback is regarding the shape of the `FormatStringUtils` API.

delphi-checks/src/test/java/au/com/integradev/delphi/utils/FormatStringUtilsTest.java
```diff
@@ -0,0 +1,212 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2023 Integrated Application Development
```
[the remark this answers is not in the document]
fourls: You can tell I started this a while ago 😄 These were the correct copyright headers when they were originally made.

delphi-checks/src/main/java/au/com/integradev/delphi/utils/FormatStringUtils.java
```diff
@@ -0,0 +1,429 @@
...
+
+    public List<DelphiFormatArgument> getArguments() {
+      if (arguments == null) {
+        List<Set<FormatSpecifierType>> typesPerArg =
+            new FormatArgumentTypeCollector(formatSpecifiers).collect();
+        List<Set<FormatSpecifier>> specifiersPerArg =
+            new FormatSpecifierCollector(formatSpecifiers).collect();
+
+        if (typesPerArg.size() != specifiersPerArg.size()) {
+          throw new IndexOutOfBoundsException(
```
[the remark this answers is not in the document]
fourls: Yeah, maybe an `IllegalStateException` would be better.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FormatParameterCountCheck.java
```diff
@@ -0,0 +1,51 @@
...
+          context,
+          arrayConstructor,
+          String.format(
+              "Add %d more formatting argument%s to this Format call.", difference, plural));
+    } else if (difference < 0) {
+      reportIssue(
+          context,
+          arrayConstructor,
+          String.format(
+              "Remove %d formatting argument%s from this Format call.", difference, plural));
```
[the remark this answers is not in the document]
fourls: Yes, oops.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FormatParameterCountCheck.java
```diff
@@ -0,0 +1,51 @@
...
+    int expectedArgs = formatString.getArguments().size();
+    int actualArgs = arrayConstructor.getElements().size();
+    int difference = expectedArgs - actualArgs;
+    String plural = Math.abs(difference) == 1 ? "" : "s";
+    if (difference > 0) {
+      reportIssue(
+          context,
+          arrayConstructor,
+          String.format(
+              "Add %d more formatting argument%s to this Format call.", difference, plural));
```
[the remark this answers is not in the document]
fourls: Sure.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FormatParameterCount.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Format parameters should match the format string",
```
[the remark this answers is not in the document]
fourls: I've replaced it with `'Format' calls should be supplied the correct number of format parameters`, what do you think?

delphi-checks/src/main/java/au/com/integradev/delphi/utils/FormatStringUtils.java
fourls: I'm happy to do this, it will be good to establish a precedent for more detailed check utilities. I also don't love utility classes that are really namespaces.

It's worth mentioning, though, that this won't ever be part of the public API, since it's in `delphi-checks`.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FormatParameterCount.json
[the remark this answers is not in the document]
Cirras: Sounds good!

delphi-checks/src/main/java/au/com/integradev/delphi/utils/FormatStringUtils.java
Cirras: > this won't ever be part of the public API

Never say never - if developers requested it in the public API, we could easily do it.
Even more easily, if we make these changes (since I'd be happy with the overall shape for a public API)

Let's do it. 👍

fourls: Very true! Reworked in my latest push.

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/NumberOrWildcard.java
```diff
@@ -0,0 +1,66 @@
...
+    if (this == o) return true;
+    if (!(o instanceof NumberOrWildcard)) return false;
+
+    NumberOrWildcard that = (NumberOrWildcard) o;
+    return (isWildcard && that.isWildcard) || (value == that.value);
+  }
+
+  @Override
+  public int hashCode() {
+    if (isWildcard) {
```
Cirras, line 60: This can just be `return Objects.hash(value, isWildcard)`, right?

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/FormatStringParser.java
```diff
@@ -0,0 +1,141 @@
...
+package au.com.integradev.delphi.utils.format;
+
+import java.text.CharacterIterator;
+import java.text.StringCharacterIterator;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class FormatStringParser {
+  private final String rawString;
```
Cirras, line 28: nit: I feel like `input` is a better name

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/FormatStringParser.java
```diff
@@ -0,0 +1,141 @@
...
+  private CharacterIterator iter;
```
Cirras, line 29: nit: I feel like the non-shortened `iterator` is better

Cirras (changes requested):
A few little nitpicks, but this is practically ready to go.

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/NumberOrWildcard.java
fourls: I didn't want wildcards that happened to have a different value be nonequal. Practically, I guess it doesn't matter, since both fields are final and you can't assign both at once.

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/FormatStringParser.java
fourls: `input` works for me. I think it's a hangover from Python that I never think to name anything `input`.

Cirras (approved):
🎉

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/NumberOrWildcard.java
```diff
@@ -0,0 +1,62 @@
...
+    return isWildcard ? Optional.empty() : Optional.of(value);
+  }
+
+  public boolean isWildcard() {
+    return isWildcard;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
```
Cirras, line 51: At most one statement is allowed per line, but 2 statements were found on this line.sonarlint(java:S122)

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/NumberOrWildcard.java
```diff
@@ -0,0 +1,62 @@
...
+    if (!(o instanceof NumberOrWildcard)) return false;
```
Cirras, line 52: `At most one statement is allowed per line, but 2 statements were found on this line.sonarlint (java:S122)`

delphi-checks/src/main/java/au/com/integradev/delphi/utils/format/NumberOrWildcard.java
```diff
@@ -0,0 +1,62 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.utils.format;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class NumberOrWildcard {
```
Cirras, line 24: `Make this class "final" or add a public constructor. sonarlint(java:S2974)`

delphi-checks/src/main/java/au/com/integradev/delphi/checks/AbstractFormatArgumentCheck.java
```diff
@@ -0,0 +1,111 @@
...
+      return;
+    }
+
+    String rawFormatString = textLiteral.get().getImageWithoutQuotes().toString();
+    FormatStringParser parser = new FormatStringParser(rawFormatString);
+    try {
+      checkFormatStringViolation(parser.parse(), arrayConstructor.get(), context);
+    } catch (DelphiFormatStringException e) {
+      handleInvalidFormatString(textLiteral.get(), context);
+      return;
```
Cirras, line 82: `Remove this redundant jump. sonarlint(java:S3626)`

delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormatArgumentCountCheckTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+class FormatArgumentCountCheckTest {
+  private DelphiTestUnitBuilder sysUtils() {
+    return new DelphiTestUnitBuilder()
+        .unitName("System.SysUtils")
+        .appendDecl("function Format(const Format: string; const Args: array of const): string;");
+  }
+
+  @Test
+  void testCorrectParametersShouldNotAddIssue() {
```
Cirras, line 33: Change `Parameters` to `Arguments` in test names.

delphi-checks/src/test/java/au/com/integradev/delphi/utils/format/FormatStringParserTest.java
```diff
@@ -0,0 +1,205 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.utils.format;
+
+import static org.assertj.core.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
Cirras, line 22: Use `assertThatThrownBy` from assertj instead.

fourls:
OK, tests renamed and all SonarLint issues resolved (hopefully). Should be good to go.

Cirras (approved):

PR #176, opened 2024-02-01
Fix symbol table executor failures on include directives with declarations and references

### LINKED ISSUE #174: Symbol table executor failure on include directives that include a symbol declaration and reference
### Prerequisites

### SonarDelphi version

1.1.0

### SonarQube version

_No response_

### Issue description

The symbol table executor currently fails with an error in some cases when executing on files containing an include directive. Specifically, if the included code both declares and uses the same symbol, the executor complains that the declaration and reference for the given symbol are overlapping.

Note that this is not a parsing error, but an error with the symbol table executor, so the other executors still run (apart from the checks executor, which depends on the symbol table executor).

### Steps to reproduce

Scan the following code.

```delphi
//////////// MyUnit.pas ////////////

unit MyUnit;

interface

{$I MyInclude.inc}

implementation

end.

////////// MyInclude.inc //////////

const
  Foo = 5;
  Bar = Foo + 10;
```

fourls:
This PR fixes #174 by excluding all symbols that are declared in an include file from the Sonar symbol table. There are two properties of symbols declared in include files that make them inappropriate to add to the Sonar symbol table:

- If more than one symbol is declared, both symbol declarations have the same text range, which is problematic as Sonar uses the declaration text range as the unique key for the symbol.
- If a symbol is both declared and referenced in the include file, an exception is raised because both the declaration and reference are overlapping.

Unfortunately this means that symbols declared in include files will not have symbol table information associated with them in the SonarQube UI, but in the grand scheme of things this isn't too much of a problem.

Note that this affects the symbol table that is reported to Sonar for use in the UI, **not** the symbol tables used by SonarDelphi itself for name resolution. Those symbol tables have full support for symbols in include files.

### DIFF
--- a/delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/scope/UnknownScopeImplTest.java
+++ b/delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/scope/UnknownScopeImplTest.java
@@ -34,6 +34,7 @@ import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.symbol.NameOccurrence;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.NameDeclaration;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.VariableNameDeclaration;
+import org.sonar.plugins.communitydelphi.api.token.DelphiToken;
 
 class UnknownScopeImplTest {
   private final UnknownScopeImpl unknownScope = UnknownScopeImpl.instance();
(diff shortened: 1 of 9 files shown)


fourls:
A subset of our internal QA checks previously exhibited the exception described in #174. I've rerun them on this branch and confirmed that they no longer appear.

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/scope/UnknownScopeImplTest.java
```diff
@@ -127,8 +128,13 @@ void testGetVariableDeclarations() {
   }
 
   private static NameOccurrence makeNameOccurrence() {
+    DelphiToken token = mock(DelphiToken.class);
```
Cirras, line 131: What's this change for?

fourls: These tests were failing because the mocked node didn't have a token, which was previously fine but now is required to determine if it's an included `SymbolicNode` or not. So I mocked a token for it.

Cirras (approved):

PR #178, opened 2024-02-01
Improve rule metadata

fourls:
This PR contains a number of miscellaneous metadata improvements:

- Rewords FormatArgumentType title to be more consistent with FormatArgumentCount
- Adds a reference to FormatArgumentType to the README
- Removes `.` from the end of the IndexLastListElement and AddressOfCharacterData rule titles
- Adds an example to IndexLastListElement
- Fix an incorrect diff and do partial rewording in FormatArgumentCount

### DIFF
--- a/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/IndexLastListElement.json
+++ b/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/IndexLastListElement.json
@@ -1,5 +1,5 @@
 {
-  "title": "The last element in a `TList` descendant should be retrieved with the `Last` method.",
+  "title": "The last element in a `TList` descendant should be retrieved with the `Last` method",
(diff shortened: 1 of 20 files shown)


delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/IndexLastListElement.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "The last element in a `TList` descendant should be retrieved with the `Last` method.",
+  "title": "The last element in a `TList` descendant should be retrieved with the `Last` method",
```
Cirras, line 2: While we're in here, may as well swap these backticks to single-quotes for consistency with the other rule titles.

Cirras (changes requested):

Cirras (approved):
Looks good!

PR #179, opened 2024-02-01
Enable on-demand plugin downloading

Cirras:
This PR adds support for the new on-demand plugin loading feature in SonarQube 10.4.

See: [The SonarScanners download only required 3rd-party plugins](https://community.sonarsource.com/t/the-sonarscanners-download-only-required-3rd-party-plugins/108156)


fourls (approved):
Looks good to me!

PR #180, opened 2024-02-05
Support ampersand-prefixed number literals

### LINKED ISSUE #173: File-level scan failure on number literal prefixed with ampersand
### Prerequisites

### SonarDelphi version

1.1.0

### SonarQube version

_No response_

### Issue description

Delphi's grammar supports the addition of an ampersand immediately before an integer or floating point literal:

```delphi
const
  MyInt = &123456;
  MyDouble = &123.456;
```

SonarDelphi currently does not support this, resulting in a scan failure when encountering the RTL unit `System.Beacon`, which uses this grammar feature seemingly pointlessly in a couple of literals.

Some observations:
* The ampersand appears to be tossed away by the compiler - it has no visible effect on the literal's value
* The ampersand must be immediately before, not separated by whitespace
    * `& 123456` causes "Identifier expected" 
* String literals do not have this behaviour
    * `&'Hello'` and `&'H'` cause "Identifier expected"
* Hexadecimal and binary literals do not have this behaviour
    * `&$1` causes "Identifier expected"
    * `$&1` causes "Missing operator or semicolon"
* Unary operators can be applied (e.g. `-&1` works, but `&-1` does not)

### Steps to reproduce

Scan the provided file, or any file that includes `System.Beacon`.

```delphi
unit MyUnit;

interface

implementation

const
  MyInt = &123456;
  MyDouble = &123.456;

begin
  WriteLn(MyInt);
  WriteLn(MyDouble);

end.
```

fourls:
Fixes #173 by accepting and discarding any ampersands before a number literal.

Some caveats:

- This occurs during parsing, not lexing (which might be more correct). This is because our ampersand handling for identifiers is in the parser. 
- This grammar will technically support whitespace between the ampersands and the number literals. I don't think this should be an issue as there's no ambiguity with any other language constructs.

### DIFF
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/DigitGroupingCheckTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/DigitGroupingCheckTest.java
@@ -36,6 +36,10 @@ class DigitGroupingCheckTest {
         "$ABC_DEF",
         "$ABC_DE_F",
         "%1_001_01",
+        "&1_000_00",
+        "&&&1_000_00",
+        "&10_00.00",
+        "&&&10_00.00",
       })
   void testNonStandardGroupingShouldAddIssue(String actual) {
     CheckVerifier.newVerifier()
@@ -62,6 +66,10 @@ class DigitGroupingCheckTest {
         "%1_01_01",
         "%10_101",
         "%10_1010",
+        "&100_000",
+        "&&&100_000",
+        "&1_000.0_0",
+        "&&&1_000.0_0",
       })
   void testStandardGroupingShouldNotAddIssue(String actual) {
     CheckVerifier.newVerifier()
--- a/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
+++ b/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
@@ -862,8 +862,6 @@ dispIDDirective              : 'dispid' expression
 // General
 //----------------------------------------------------------------------------
 ident                        : TkIdentifier
-                             | '&'! identifierOrKeyword
-                             | '&' '&' identifierOrKeyword -> ^({combineLastNTokens(2)})
                              | keywordsUsedAsNames -> ^({changeTokenType(TkIdentifier)})
                              ;
 identifierOrKeyword          : TkIdentifier
@@ -1090,7 +1088,7 @@ DEREFERENCE          : '^'  ;
 ADDRESS              : '@'  ;
 DOT                  : '.'  ;
 DOT_DOT              : '..' ;
-AMPERSAND            : '&'  ;
+AMPERSAND            : '@AMPERSAND@'  ;
 
 //****************************
 // Imaginary tokens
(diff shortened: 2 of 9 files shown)


Cirras (changes requested):
> This occurs during parsing, not lexing (which might be more correct). This is because our ampersand handling for identifiers is in the parser.

I'd strongly prefer handling this in the lexer.
We can make that same change to identifiers, unless you think there's a good reason not to.

This would also involve deprecating the `AMPERSAND` token in the grammar (just change the value from `'&'` to `'@AMPERSAND@'`.

I'll come along later and start generating deprecations for `DelphiTokenType` enum values based on some sort of token naming conventions, but we can leave that for now.

fourls:
I have:
- Moved the handling of ampersands for identifiers, int literals, and real literals down into the lexer
- Overridden `IdentifierNodeImpl.getImage` to strip up to one leading ampersand
- Modified `IntLiteralNodeImpl.getValue` to handle ampersands in the image
- Modified `RealLiteralNodeImpl.getValue` to handle ampersands in the image
- Modified `DigitGroupingCheck` to handle ampersands in literal images

This seems to work. I've added some extra testing, and the existing symbol table test at `DelphiSymbolTableExecutorTest.testAmpersands` passes.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/DigitGroupingCheckTest.java
```diff
@@ -36,6 +36,10 @@ class DigitGroupingCheckTest {
         "$ABC_DEF",
         "$ABC_DE_F",
         "%1_001_01",
+        "&1_001_01",
```
Cirras, line 39: nit: these look like a binary literal at first, can you change the non-leading 1s to 0?

delphi-checks/src/test/java/au/com/integradev/delphi/checks/DigitGroupingCheckTest.java
```diff
@@ -62,6 +66,10 @@ void testNonStandardGroupingShouldAddIssue(String actual) {
         "%1_01_01",
         "%10_101",
         "%10_1010",
+        "&100_101",
```
Cirras, line 69: nit: these look like a binary literal at first, can you change the non-leading 1s to 0?

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -81,6 +81,15 @@ package au.com.integradev.delphi.antlr;
     return "line " + e.line + ":" + e.charPositionInLine;
   }
 
+  private boolean nextNonAmpersandIsDigit() {
```
Cirras, line 84: Unused function.

delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/GrammarTest.java
```diff
@@ -262,6 +262,11 @@ void testUnusualIntegerLiterals() {
     assertParsed("UnusualIntegerLiterals.pas");
   }
 
+  @Test
+  void testAmpersandNumberLiterals() {
```
Cirras, line 266: I feel like we normally say "numeric literals" instead of "number literals".

Cirras (changes requested):

Cirras (approved):
Well done! 👍

PR #185, opened 2024-02-09
Remove dependency on `battila7/get-version-action`

Cirras:
This PR fixes warnings in the release workflow by removing our dependency on `battila7/get-version-action`.

The `get-version-action` action has a couple of issues:
- It uses the deprecated `set-output` (https://github.com/battila7/get-version-action/issues/22)
- It runs on the deprecated Node 12 (https://github.com/battila7/get-version-action/issues/18)

This action doesn't do much though, and our use-case can be trivially satisfied using `GITHUB_REF_NAME`.


zaneduffield (approved):
LGTM 👍

PR #187, opened 2024-02-12
Bump Orchestrator to v4.7.1.1872

Cirras:
SonarSource has released a patch to fix an issue where Orchestrator was pulling the wrong SonarQube version for the `LATEST_RELEASE` alias (which we use).


zaneduffield (approved):
LGTM 👍

PR #190, opened 2024-02-26
Make imaginary tokens unmatchable

Cirras:
This is a silly mistake that we inherited in the Delphi grammar.

We've been defining our imaginary tokens with actual values, and not thinking much of it. However, this technically means that they could be matched in source code by the lexer and wreak all sorts of havoc.

Declaring imaginary tokens in a `tokens` block is a clean and easy way of preventing them from ever being matched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Static char arrays weren't accepted for `'%s'` in `FormatArgumentType`.
+- Imaginary tokens could be erroneously matched by the lexer in valid source code.
 
 ## [1.2.0] - 2024-02-02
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Static char arrays weren't accepted for `'%s'` in `FormatArgumentType`.
+- Obscure grammar bug where imaginary tokens could technically be matched in valid source code.
```
fourls, line 22: It might be better to keep this entry in the same format as the others?
It is true it was obscure, but it's not really relevant to the change itself.

Thoughts on something like "Lexer incorrectly identifying imaginary tokens in valid source code."?

fourls (commented):
If it's good enough for the unit tests, it's good enough for me! All looks good.
Just some thoughts about the changelog.

fourls (approved):
Looks great!

PR #196, opened 2024-03-01
Fix null reference in PublicField when encountering anonymous records

fourls:
This PR fixes a null pointer exception that occurs when `PublicFieldCheck` encounters an anonymous record. 

I've also bundled in the removal of a dead rule from the grammar.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PublicFieldCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PublicFieldCheck.java
@@ -24,7 +24,7 @@ package au.com.integradev.delphi.checks;
 
 import org.sonar.check.Rule;
 import org.sonar.plugins.communitydelphi.api.ast.FieldDeclarationNode;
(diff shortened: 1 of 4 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/PublicFieldCheck.java
```diff
@@ -43,6 +43,8 @@ public DelphiCheckContext visit(FieldDeclarationNode field, DelphiCheckContext c
   }
 
   private static boolean isRecordField(FieldDeclarationNode field) {
-    return field.getFirstParentOfType(TypeDeclarationNode.class).isRecord();
+    TypeDeclarationNode typeParent = field.getFirstParentOfType(TypeDeclarationNode.class);
```
Cirras, line 46: Do `TypeNode` instead and you won't need to do any null-checking - it should always be present.

Cirras (changes requested):

Cirras (approved):
Looks good to me.

PR #197, opened 2024-03-01
Gracefully handle exceptions in check execution

fourls:
Currently, the entire checks executor crashes when a single check has an exception raised. This results in nondeterministic issues being raised for files that cause certain checks to crash, which is not ideal.

This PR handles exceptions in check execution by moving onto the next check instead.


Cirras (approved):
👍

PR #198, opened 2024-03-04
Remove Orchestrator proxy settings

Cirras:
We don't need proxy settings for our integration tests, since the Orchestrator library now downloads artifacts through maven central.


Cirras:
Actually, looks like these are still needed.
I wonder why it worked in my environment...

PR #199, opened 2024-03-04
Speed up the analyzer

Cirras:
This PR improves the analyzer's performance by ~30% in internal testing.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
@@ -71,6 +71,7 @@ public class DelphiPreprocessor {
 
   private DelphiTokenStream tokenStream;
   private Set<Token> tokens;
+  private List<Token> rawTokens;
   private int tokenIndex;
 
   DelphiPreprocessor(DelphiLexer lexer, DelphiFileConfig config, Platform platform) {
@@ -119,6 +120,7 @@ public class DelphiPreprocessor {
 
     tokenStream.fill();
     tokens = extractTokens(tokenStream);
+    rawTokens = List.copyOf(tokens);
     tokens.forEach(this::processToken);
     directives.stream()
         .map(CompilerDirectiveImpl.class::cast)
@@ -317,6 +319,13 @@ public class DelphiPreprocessor {
 
   public TypeFactory getTypeFactory() {
     return config.getTypeFactory();
+  }
+
+  public List<DelphiToken> getRawTokens() {
+    return rawTokens.stream()
+        .filter(token -> token.getType() != Token.EOF)
+        .map(DelphiTokenImpl::new)
+        .collect(Collectors.toUnmodifiableList());
   }
 
   static class SelfReferencingIncludeFileException extends RuntimeException {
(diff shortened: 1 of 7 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
```diff
@@ -119,6 +121,7 @@ public void process() {
 
     tokenStream.fill();
     tokens = extractTokens(tokenStream);
+    rawTokens = List.copyOf(tokens);
```
fourls, line 124: It feels unnecessary to have the tokens stored as a sorted TreeSet, when they end up being stored in an ArrayList anyway.

Retooling `extractTokens` to return a `List<Token>` could have the benefit of slightly more idiomatic code, in which:

- `tokens` is explicitly ordered
- `tokens` is a copy of `rawTokens`, rather than the other way around

```java
private List<Token> tokens;
private List<Token> rawTokens;

// ...

DelphiProcessor(DelphiLexer lexer, DelphiFileConfig config, Platform platform) {
  // ...
  rawTokens = extractTokens(tokenStream);
  tokens = List.copyOf(rawTokens);
  // ...
  tokenStream.setTokens(tokens);
  // ...
}
```

fourls (changes requested):
Looks great! These performance improvements are sizeable.

I've left a few (minor) comments. I'll also run this branch through our QA checks for more assurance that there haven't been any accidental behaviour changes.

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
Cirras: How would you handle including new tokens into the `tokens` list via include directives with this approach?

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
```diff
@@ -319,6 +322,13 @@ public TypeFactory getTypeFactory() {
     return config.getTypeFactory();
   }
 
+  public List<DelphiToken> getRawTokens() {
+    return rawTokens.stream()
+        .map(DelphiTokenImpl::new)
+        .filter(Predicate.not(DelphiToken::isEof))
```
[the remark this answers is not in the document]
Cirras: Sure, we may as well avoid an unnecessary object creation on `EOF`.

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/DelphiPreprocessor.java
fourls: Discussed in person - performance of ArrayList<Token> would be worse. Resolving 👍

fourls:
Our QA results show a little bit of churn, but they seem to be mostly correct (or not overtly incorrect, at least).

Some observations:

- Removed some `EmptyBlock` and `EmptyRoutine` false positive issues where a supplied comment should have already suppressed the issue
- Fixed file pointer errors when reading files in unusual encodings
- 5-10 cases of unexplained issue churn - adding/removing a few isolated `UnusedImport`s, removing an `InheritedMethodWithNoCode` on a destructor that just calls `inherited`

It seems that the analysis results have improved, although the unexplained churn seems unusual - I'm not sure if those are correct or not.

fourls (approved):
These changes look good. The confusing aspects of the churn are likely related to #200 and other existing bugs, not this change - although they may have been exposed by changes in the token list storage.

PR #200, opened 2024-03-05
Fix InheritedMethodWithNoCode incorrectly retrieving class declarations

fourls:
When traversing a routine's inheritance tree, InheritedMethodNoCode currently does not distinguish between class routines and member routines. This can cause the wrong routine to be chosen if a class routine with the same name and number of parameters exists in the inheritance tree.

This PR fixes this issue by filtering out non-callable routines.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/InheritedMethodWithNoCodeCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/InheritedMethodWithNoCodeCheck.java
@@ -42,6 +42,7 @@ import org.sonar.plugins.communitydelphi.api.ast.utils.ExpressionNodeUtils;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.RoutineDirective;
+import org.sonar.plugins.communitydelphi.api.symbol.declaration.RoutineKind;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.RoutineNameDeclaration;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.TypeNameDeclaration;
 import org.sonar.plugins.communitydelphi.api.type.Type;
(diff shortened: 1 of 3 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/InheritedMethodWithNoCodeCheck.java
```diff
@@ -195,6 +195,7 @@ private static boolean argumentSignaturesMatch(
   private static boolean isOverriddenMethod(
       RoutineNameDeclaration parent, RoutineNameDeclaration child) {
     return parent.getName().equalsIgnoreCase(child.getName())
-        && parent.getParameters().equals(child.getParameters());
+        && parent.getParameters().equals(child.getParameters())
+        && parent.isCallable();
```
Cirras, line 199: Class methods have polymorphism in Delphi and can call `inherited` too.
```suggestion
        && parent.isClassInvocable() == child.isClassInvocable()
        && parent.isCallable();
```

Cirras (changes requested):
Looks good, just want one additional condition to disambiguate class methods.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/InheritedMethodWithNoCodeCheck.java
fourls: Unfortunately, `inherited` on an instance routine can refer to a class routine in the parent:

```pascal
type
  TBase = class(TObject)
  protected
    class procedure MyProc;
  end;

  TChild = class(TBase)
  public
    procedure MyProc;
  end;

procedure TChild.MyProc;
begin
  Writeln('instance proc');
  inherited MyProc;
end;

class procedure TBase.MyProc;
begin
  Writeln('class proc');
end;

initialization
  TChild.Create.MyProc; // writes 'instance proc\nclass proc'
```

fourls: I've implemented this logic and added some tests to confirm.

Cirras (approved):

PR #202, opened 2024-03-15
Full support for Delphi 12

Cirras:
This PR finishes implementing support for Delphi 12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for the `TEXTBLOCK` directive.
 - **API:** `CompilerDirectiveParser` can now return a new `TextBlockDirective` type.
+- **API:** `CheckVerifier::withCompilerVersion` method.
+- **API:** `CheckVerifier::withToolchain` method.
 
 ### Changed
 
+- `NativeInt` and `NativeUInt` are now treated as weak aliases in Delphi 12+.
 - Performance improvements.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for multiline string literals within compiler directives.
 - Support for the `TEXTBLOCK` directive.
 - **API:** `CompilerDirectiveParser` can now return a new `TextBlockDirective` type.
 - **API:** `CheckVerifier::withCompilerVersion` method.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for the `TEXTBLOCK` directive.
+- **API:** `CompilerDirectiveParser` can now return a new `TextBlockDirective` type.
+
 ### Changed
 
 - Performance improvements.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `NativeInt` and `NativeUInt` are now treated as weak aliases in Delphi 12+.
+- The length of open arrays is now modeled as `NativeInt` in Delphi 12+.
 - Performance improvements.
 
 ### Fixed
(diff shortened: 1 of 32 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java
```diff
@@ -89,11 +93,37 @@ public static Type argumentByIndex(int index) {
...
+    private final Type nativeIntType;
+
+    private LengthReturnType(TypeFactory typeFactory) {
+      this.byteType = typeFactory.getIntrinsic(IntrinsicType.BYTE);
+      this.integerType = typeFactory.getIntrinsic(IntrinsicType.INTEGER);
+      this.nativeIntType = typeFactory.getIntrinsic(IntrinsicType.NATIVEINT);
+    }
+
+    @Override
+    public Type getReturnType(List<Type> arguments) {
```
fourls, line 108: This is Delphi 12-only behaviour, but it's being applied for all versions. Is this intentional?

CHANGELOG.md
```diff
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for multiline string literals within compiler directives.
+- Support for the `TEXTBLOCK` directive.
+- **API:** `CompilerDirectiveParser` can now return a new `TextBlockDirective` type.
+- **API:** `CheckVerifier::withCompilerVersion` method.
+- **API:** `CheckVerifier::withToolchain` method.
```
fourls, line 16: Include new API method `DelphiFile::getTextBlockLineEndingModeRegistry`.

CHANGELOG.md
```diff
@@ -7,8 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed
 
+- `NativeInt` and `NativeUInt` are now treated as weak aliases in Delphi 12+.
+- The length of open arrays is now modeled as `NativeInt` in Delphi 12+.
```
fourls, line 21: The length of open arrays is now modeled as `NativeInt` regardless of Delphi version. If this is unintentional, it should be fixed elsewhere, but if it's intended behaviour then this changelog entry should be changed.

fourls (changes requested):
Sorry for the delay in reviewing - looks great for the most part! Just a couple of questions and a changelog improvement.

There's been a few changes in master since this branch forked, so would be good to get it rebased on master as well.

CHANGELOG.md
Cirras: I actually don't want `DelphiAst::getDelphiFile` to be exposed in the API because `DelphiFile` isn't in the custom rules API. Its API surface has a lot of overlap with `DelphiCheckContext`, and it feels quite inappropriate to go to the root of the AST to get some additional metadata about the source file.

Instead of doing this, I'll be deprecating `DelphiAst::getDelphiFile`.

CHANGELOG.md
Cirras: Unintended behavior, the changelog is correct.

delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java
Cirras: Nope! Good catch.

fourls:
Code change looks good, but it seems to have significantly increased the execution time of the `delphi-checks` tests - on GitHub Actions it's doubled, going from [4m 50s](https://github.com/integrated-application-development/sonar-delphi/actions/runs/8385296401/job/22963974269) in master to [9m 46s](https://github.com/integrated-application-development/sonar-delphi/actions/runs/8430119706) on this branch. Ideally we'd get this back down to a similar amount of time before merging.

Initial analysis and profiling suggests that it's due to an uptick in the volume of runtime classes generated for weak alias types.

Cirras:
> Code change looks good, but it seems to have significantly increased the execution time of the `delphi-checks` tests - on GitHub Actions it's doubled, going from [4m 50s](https://github.com/integrated-application-development/sonar-delphi/actions/runs/8385296401/job/22963974269) in master to [9m 46s](https://github.com/integrated-application-development/sonar-delphi/actions/runs/8430119706/job/23085459812) on this branch. Ideally we'd get this back down to a similar amount of time before merging.
> 
> Initial analysis and profiling suggests that it's due to an uptick in the volume of runtime classes generated for weak alias types.

This was because we started generating weak aliases as part of `TypeFactoryImpl` construction.
We used to cache type alias classes per `TypeAliasGenerator` - that cache is now static and the classes are generated only once.

fourls (approved):
All looks good now! Tests are back down to usual durations (perhaps even slightly faster).
Looks like we can close the book on Delphi 12 compatibility now 😄

PR #203, opened 2024-03-15
Fix type resolution inaccuracies around subtract expressions

Cirras:
By random chance, I noticed that `SubtractDataAggregator` wasn't actually being used in `OperatorOverloadResolutionTest`

This PR fixes the test and implements the correct type resolution for subtract expressions.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
@@ -258,7 +258,7 @@ public class OperatorInvocableCollector {
       case ADD:
         return createIntegerArithmeticBinary("Add");
       case SUBTRACT:
-        return createIntegerArithmeticBinary("Subtract");
+        return createSubtract();
       case MULTIPLY:
         return createIntegerArithmeticBinary("Multiply");
       case DIVIDE:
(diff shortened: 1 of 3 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
```diff
@@ -439,6 +439,38 @@ private Set<Invocable> createIntegerArithmeticBinary(String name) {
     return result;
   }
 
+  private Set<Invocable> createSubtract() {
```
zaneduffield, line 442: This only differs from `createIntegerArithmeticBinary` in the result type for `uint31 <op> uint31`.
I'd say this should be reflected in the code (e.g. by pulling out the common code into a function that accepts the return type for the differing case), unless you have a particular reason for wanting the code to be flat.

zaneduffield (commented):
Damn! I thought I made sure the tests were checking the right thing when I reviewed this the first time...
Oh well, I'm glad you found it.

The fix looks good, just one comment about code duplication.

delphi-frontend/src/main/java/au/com/integradev/delphi/operator/OperatorInvocableCollector.java
Cirras: I can't think of an obvious way to abstract this logically, do you have any thoughts?

zaneduffield: My idea was basically just to have the return type of the `uint31 <op> uint31` intrinsic be a parameter to a the function which wraps the common code. Then you could have `createSubtract` just call that function and provide the output type for that intrinsic operator.

Cirras: Done.

zaneduffield (approved):
LGTM 👍

PR #208, opened 2024-03-22
Handle UNC paths correctly during analysis

fourls:
This PR fixes a bug that causes UNC paths to be mangled when converting the Sonar-provided file URI to an absolute path.

`java.io.File` and `java.nio.file.Path` are both convertible to a `java.net.URI`, but they treat UNC paths differently:

- `File` places the entire path in the path portion of the URI - `\\host\a\b\c` becomes `file:////host/a/b/c`
- `Path` (correctly) places the host in the host portion of the URI - `\\host\a\b\c` becomes `file://host/a/b/c`

We were mixing and matching our approach by feeding the `File` constructor with a URI constructed from a `Path` object. This PR fixes the issue by creating a `Path` from the URI, then creating the file via `Path.toFile` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Type resolution inaccuracies around subtract expressions.
 - The wrong inherited method could be found in `InheritedMethodWithNoCode`, causing false negatives.
+- Exception when scanning UNC paths.
 
 ## [1.3.0] - 2024-03-01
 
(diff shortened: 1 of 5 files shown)


fourls:
Note that to test this, you have to call the `sonar-scanner` jar directly using `java` - because it uses CMD, the `sonar-scanner.bat` startup script does not work with UNC paths.

Cirras (changes requested):
Nice detective work!
Also delighted to see the `uriToAbsolutePath` util removed.

A couple minor suggested changes, since this `Path.get(uri)` pattern is already wrapped up in an `inputFileToPath` util.

fourls:
Ah, didn't notice that method. I've now updated to use it.

Cirras (approved):
Looks good, thanks. 👍

PR #219, opened 2024-04-22
Fix cast exception when parsing fully qualified attributes

### LINKED ISSUE #213: ClassCastException on qualified attribute references
### Prerequisites

### SonarDelphi version

1.4.0

### SonarQube version

10.4.1

### Issue description

The following error occurs on source code containing fully qualified attribute references.
```
java.lang.ClassCastException: class au.com.integradev.delphi.symbol.occurrence.NameOccurrenceImpl cannot be cast to class au.com.integradev.delphi.symbol.occurrence.AttributeNameOccurrenceImpl (au.com.integradev.delphi.symbol.occurrence.NameOccurrenceImpl and au.com.integradev.delphi.symbol.occurrence.AttributeNameOccurrenceImpl are in unnamed module of loader org.sonar.classloader.ClassRealm @57abad67
```

### Steps to reproduce

Scan the following code example.

```pas
uses
  REST.Json.Types, Classes;

type
  TTest = class(TInterfacedPersistent)
  private
    [REST.Json.Types.JSONMarshalledAttribute(False)]
    FTest: Integer;
  public
    property Test: Integer read FTest write FTest;
  end;
```

fourls:
This PR fixes a `ClassCastException` that occurs when parsing fully qualified attribute references.

Fixes #213

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
@@ -572,7 +572,7 @@ public class NameResolver {
     }
   }
 
-  private void readNameReference(NameReferenceNode node, boolean attribute) {
+  private void readNameReference(NameReferenceNode node, boolean inAttribute) {
     boolean couldBeUnitNameReference =
         currentScope == null
             || (!(currentScope instanceof UnknownScope) && currentScope.equals(node.getScope()));
@@ -582,12 +582,12 @@ public class NameResolver {
         return;
       }
 
-      NameOccurrence occurrence = createNameOccurrence(reference, attribute);
+      NameOccurrence occurrence = createNameOccurrence(reference, inAttribute);
       resolveNameReferenceOccurrence(occurrence, reference.nextName() == null);
 
       if (nameResolutionFailed()) {
         if (couldBeUnitNameReference) {
-          readPossibleUnitNameReference(node);
+          readPossibleUnitNameReference(node, inAttribute);
         }
         return;
       }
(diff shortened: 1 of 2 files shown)


zaneduffield:
We should probably add a test that covers this case, to prevent regressions.

fourls:
I agree - happy to hear recommendations about how to test this change. Usually these sorts of things are in `DelphiSymbolExecutorTest`, but the API for these tests doesn't let you assert occurrences across units, so it would need a change to the way these tests work.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
```diff
@@ -664,9 +664,9 @@ private static boolean isDynamicArrayReference(Type type) {
     return type.isClassReference() && ((ClassReferenceType) type).classType().isDynamicArray();
   }
 
-  private void readPossibleUnitNameReference(NameReferenceNode node) {
+  private void readPossibleUnitNameReference(NameReferenceNode node, boolean inAttribute) {
```
Cirras, line 667: Let's call it `attribute` to be consistent with the parameter name in `readNameReference`.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
fourls: I actually regret naming that parameter `attribute`, I feel like it's an unexpected name for a boolean to have. I've changed it instead to be consistently `inAttribute`, what do you think of that?

Cirras: Fine by me.

Cirras (approved):
Looks good to me!

PR #223, opened 2024-04-29
Improve usage of local scopes in the symbol table

### LINKED ISSUE #220: `DuplicatedDeclarationException` on inline variables in a `try` statement
### Prerequisites

### SonarDelphi version

1.4.0

### SonarQube version

10.4.1

### Issue description

A `DuplicatedDeclarationException` exception occurs while parsing simple Delphi code where the same variable name is used twice in a `try..finally` or `try..except`. Once in the `try` and once in the `finally` or `except` block.

### Steps to reproduce

To reproduce this, you declare a variable in the `try`-block and use the same variablename in the `except`- or `finally`-block.

``` delphi
procedure TMyClass.ThisCodeFails;
begin
  try
    var LVar := 'test';
  finally
    var LVar := 'test2';
  end;
end;

procedure TMyClass.ThisFailsToo;
begin
  try
    var LVar := 'test';
  except
    var LVar := 'test2';
  end;
end;
```

Cirras:
This PR improves the modeling of local scopes in the symbol table, which fixes #220.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,14 +22,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - "Separate grouped parameters" quick fix for `GroupedParameterDeclaration`.
 - "Remove (n) unused formatting arguments" quick fix for `FormatArgumentCount`.
 - "Use string value directly" quick fix for `FormatArgumentCount`.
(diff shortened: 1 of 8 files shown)


fourls (changes requested):
Looks good, just one minor comment.

PR #225, opened 2024-05-01
Fix if guard expressions being incorrectly resolved

fourls:
This PR fixes a regression introduced by #223, which causes `if` guard expressions to be skipped over by the name resolver. 

This was manifesting in QA as large numbers of erroneous UnusedRoutine, UnusedLocalVariable, and VariableInitialization issues, and I have re-run a subset of the QA checks on this PR and found these issues are no longer cropping up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Exception when parsing fully qualified attribute references.
 - `DuplicatedDeclarationException` errors caused by some local scopes being modeled incorrectly.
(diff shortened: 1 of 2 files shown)


fourls:
Some other QA checks raised unusual new issues that led me to a pre-existing symbol table generation bug.

Some statements (`if`, `for`, and `with`) correctly created a local scope, but then visited the *children* of their body statement instead of the body statement itself. This was previously OK in most cases, since there wasn't specialised handling, but with the changes in #223 it was causing unusual behaviour.

To help prevent this issue in future, I've renamed a number of methods to make their behaviour clearer.

Cirras (approved):
Great work on finding this subtle bug.
All looks good to me. 👍

PR #229, opened 2024-05-07
Correctly handle uppercase format specifiers in format strings

### LINKED ISSUE #228: Wrong report if `SysUtils.Format` has parameter with Upper-cased TYPE.
### Prerequisites

### SonarDelphi version

1.5.0

### SonarQube version

10.4

### Issue description

Using `SysUtils.Format` with Upper-cased TYPE reports wrong detection.
It was reported as "Fix this invalid format string.".

TYPE may be specified in uppercase as well as in lowercase; both produce the same results.
So this code has no problem.

It seems other TYPE in uppercase reports same wrong detection.

[https://docwiki.embarcadero.com/Libraries/Sydney/en/System.SysUtils.Format#Format_Strings](url)

### Steps to reproduce

Analysing the attached Delphi unit using Sonar-Delphi triggers this issue.

~~~
function GetStr: string;
var
  ToBeShown: Integer;
begin
  ToBeShown := $12;
  Result := Format('%.4X', [ToBeShown]);  // this makes '0012'.  simular to "%.4x"
end;
~~~

fourls:
This PR modifies the format string parser to match format specifiers case-insensitively, as specified in the Delphi documentation.

Fixes #228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.5.0] - 2024-05-02
 
+### Fixed
+
+- False positive `FormatStringValid` issues on strings containing uppercase format specifiers.
+
 ### Added
 
 - Support for the `winapi` calling convention.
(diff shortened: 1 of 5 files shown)


Cirras (approved):
Looks good!

PR #230, opened 2024-05-10
Correct ordering of GroupedParameterDeclaration quick fix

fourls:
This PR fixes a sequencing bug in the GroupedParameterDeclaration quick fix, in which text edits would be applied out of order.


Cirras (approved):
Looks good. 👍

PR #235, opened 2024-05-16
Exclude `@SonarLintUnsupported` rules in a SonarLint context

fourls:
This PR excludes checks annotated with `@SonarLintUnsupported` from being registered in the rule repository (i.e. in `DelphiRulesDefinition`. This fixes unsupported rules being defined and reported in a SonarLint context.

### DIFF
--- a/sonar-delphi-plugin/src/main/java/au/com/integradev/delphi/DelphiRulesDefinition.java
+++ b/sonar-delphi-plugin/src/main/java/au/com/integradev/delphi/DelphiRulesDefinition.java
@@ -20,10 +20,15 @@ package au.com.integradev.delphi;
 
 import au.com.integradev.delphi.checks.CheckList;
 import au.com.integradev.delphi.core.Delphi;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.sonar.api.SonarProduct;
 import org.sonar.api.SonarRuntime;
 import org.sonar.api.server.rule.RulesDefinition;
+import org.sonar.api.utils.AnnotationUtils;
 import org.sonar.plugins.communitydelphi.api.check.MetadataResourcePath;
 import org.sonar.plugins.communitydelphi.api.check.RuleTemplateAnnotationReader;
+import org.sonar.plugins.communitydelphi.api.check.SonarLintUnsupported;
 import org.sonarsource.analyzer.commons.RuleMetadataLoader;
 
 public class DelphiRulesDefinition implements RulesDefinition {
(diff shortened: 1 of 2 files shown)


sonar-delphi-plugin/src/main/java/au/com/integradev/delphi/DelphiRulesDefinition.java
```diff
@@ -53,9 +58,19 @@ public void define(Context context) {
...
+
+    ruleMetadataLoader.addRulesByAnnotatedClass(repository, checks);
+    ruleTemplateAnnotationReader.updateRulesByAnnotatedClass(repository, checks);
 
     repository.done();
   }
+
+  private boolean isCheckSupportedOnPlatform(Class<?> check) {
+    return runtime.getProduct() == SonarProduct.SONARQUBE
+        || AnnotationUtils.getAnnotation(check, SonarLintUnsupported.class) == null;
```
Cirras, line 74: This is more explicit and targeted to the sonarlint context.
```suggestion
    return !(runtime.getProduct() == SonarProduct.SONARLINT
        && AnnotationUtils.getAnnotation(check, SonarLintUnsupported.class) != null);
```

Cirras (changes requested):
One comment for now, but I have concerns around the brittleness of the approach (in that we'd be expecting any custom rules plugins to remember to do this in their rules definition)

sonar-delphi-plugin/src/main/java/au/com/integradev/delphi/DelphiRulesDefinition.java
fourls: That seems much more confusing to me - instead of having no negations, it has two at different levels.

I copied the approach in DelphiChecksExecutor - see:

https://github.com/integrated-application-development/sonar-delphi/blob/e4d61b6479e86e177b4ac149e6305e9d7292e844/delphi-frontend/src/main/java/au/com/integradev/delphi/executor/DelphiChecksExecutor.java#L102-L105

fourls:
I don't know if brittle is the right word - if custom rules plugins report "here are my rules!" in a SonarLint context and those rules aren't actually available, then I would probably lean towards saying that's on them. There are also no practical consequences for a custom rules plugin reporting rules that won't be executed.

sonar-delphi-plugin/src/main/java/au/com/integradev/delphi/DelphiRulesDefinition.java
fourls: I suppose the way you suggest is more resilient to future additions to the `SonarProduct` enum. Should I change DelphiChecksExecutor as well?

Cirras: Huh. I'm surprised that I wrote it that way in the checks executor...
I think it's fine, let's just leave it as-is for now.
A new `SonarProduct` with special rules would be a big change regardless and prompt us to review `SonarProduct` usages anyway.

Cirras:
> I don't know if brittle is the right word - if custom rules plugins report "here are my rules!" in a SonarLint context and those rules aren't actually available, then I would probably lean towards saying that's on them. There are also no practical consequences for a custom rules plugin reporting rules that won't be executed.
> 
> What this change makes sure is that we, the core plugin, are correctly reporting the rules that are actually available in a SonarLint context.

Hard to articulate my problem here, I might just be annoyed that there's no source of truth for the reporting behavior even though we've gone out of our way to provide this nice annotation that prevents the core analyzer from actually **executing** the rule.

Previously the `@SonarLintUnsupported` annotation was handled internal to the core plugin and custom ruels plugins got to just consume them and tag rules with them.
With this change, we'd be expecting them to also handle them in their rule definitions or they'll be doing *subtly* wrong behavior and reporting rules to DelphiLint that won't actually be executed by DelphiLint.

With that being said, I don't see any obvious way to avoid it. 😩
Might give it some more thought and hope I have an epiphany.

fourls:
If you're worried about DelphiLint, custom rules plugins can't be used in standalone mode so this wouldn't ever surface as an actual inconsistency. While I suppose this could *theoretically* be a future DelphiLint feature, I don't think I would ever implement it unless given a substantial reason why it's helpful.

I do appreciate that we're giving them this magic annotation that actually doesn't do all the magic anymore. That is unfortunate.

Cirras:
> custom rules plugins can't be used in standalone mode so this wouldn't ever surface as an actual inconsistency

Oh hey, you got me there. 👀
That probably should have been obvious to me, I'd say it's not a serious concern then.

Cirras (changes requested):
Mind adding a changelog entry?

PR #239, opened 2024-05-20
Ensure BOM is not present in `DelphiCheckContext::getFileLines`

### LINKED ISSUE #234: File pointer errors on the first line of files where BOM is present
### Prerequisites

### SonarDelphi version

1.5.0

### SonarQube version

10.5.1

### Issue description

When making a complete unit deprecated with a short justification, the first line of the unit might grow over 120 chars quite fast. SonarDelphi fails in this situation though with an error message that some index is not a valid index.

### Steps to reproduce

1. Create a new unit 'MyDeprecatedTestUnit.pas' in a project 
2. Declare the unit as deprecated with a justification/steps to resolve deprecated warnings:

```
unit MyDeprecatedTestUnit deprecated 'The mechanism in this unit has been replaced with another one, see MyNewApproach.ps -oSomeUser -c05/15/2024';
interface
const
implementation
end.
```

3. Note that the project compiles, showing a deprecated warning if the unit is referenced
4. Ensure the "Lines should not be too long" rule is active and configured to a limit of 120 (seems both to be default)
5. Analyze the project
6. Exception during the parsing

_No response_

Cirras:
This PR fixes file pointer errors caused by an inconsistency in the 2 ways source files are read during `DelphiFile` construction.
* `DelphiFileStream` recognizes BOMs and strips them before passing the string data forward to the preprocessor
* `FileUtils.readLines` has no handling for BOMs and persists them into the string data

This inconsistency means that the length of the first line is incorrect according to `DelphiCheckContext::getFileLines`, and any rule relying on this API may accidentally report issues with positions that run over the end of the line. 

Fixes #234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- File pointer errors when an issue is raised on the first line of a file with a byte order mark.
 - Symbol table construction errors for program files omitting the `program` header.
 - Parsing errors on parenthesised anonymous methods.
 - Incorrect ordering of edits in the "Separate grouped parameters" quick fix for `GroupedParameterDeclaration`.
--- a/delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/AbstractDelphiTestFile.java
+++ b/delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/AbstractDelphiTestFile.java
@@ -55,6 +55,8 @@ abstract class AbstractDelphiTestFile<T extends AbstractDelphiTestFile<T>>
       file.deleteOnExit();
 
       try (FileWriter fileWriter = new FileWriter(file, UTF_8)) {
+        // Prepend UTF-8 BOM
+        fileWriter.write('\ufeff');
         fileWriter.write(sourceCode());
         fileWriter.flush();
       }
--- a/delphi-frontend/src/test/java/au/com/integradev/delphi/file/DelphiFileTest.java
+++ b/delphi-frontend/src/test/java/au/com/integradev/delphi/file/DelphiFileTest.java
@@ -71,7 +71,7 @@ class DelphiFileTest {
             Collections.emptySet());
 
     DelphiFile delphiFile = DelphiInputFile.from(inputFile, config);
-    assertThat(delphiFile.getSourceCodeFilesLines().get(4)).isEqualTo("// €†šŸÀÿ");
+    assertThat(delphiFile.getSourceCodeFileLines().get(4)).isEqualTo("// €†šŸÀÿ");
   }
 
   @Test
@@ -87,6 +87,6 @@ class DelphiFileTest {
             Collections.emptySet());
 
     DelphiFile delphiFile = DelphiFile.from(file, config);
-    assertThat(delphiFile.getSourceCodeFilesLines().get(4)).hasSize(120);
+    assertThat(delphiFile.getSourceCodeFileLines().get(4)).hasSize(120);
   }
 }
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- File pointer errors on issues on the first line of files where BOM is present.
```
fourls, line 20: This reads a little stilted - maybe something like this?

```suggestion
- File pointer errors when an issue is raised on the first line of a file with a byte order mark.
```

delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/AbstractDelphiTestFile.java
```diff
@@ -55,6 +55,7 @@ protected InputFile inputFile() {
       file.deleteOnExit();
 
       try (FileWriter fileWriter = new FileWriter(file, UTF_8)) {
+        fileWriter.write('\ufeff');
```
fourls, line 58: Do you think it could be worth adding a quick explanatory comment here?

delphi-frontend/src/test/java/au/com/integradev/delphi/file/DelphiFileTest.java
fourls, line 1: It might be good to add a test asserting that `getSourceCodeFileLines()` does not include the BOM, e.g. something like this:

```java
String firstLine = delphiFile.getSourceCodeFileLines().get(0);
assertThat(firstLine).doesNotStartWith("\ufeff");
```

fourls (changes requested):
Code change looks good. Just a couple of minor thoughts.

CHANGELOG.md
Cirras: Sure, I like the suggested version so I'll just use that directly.

delphi-frontend/src/test/java/au/com/integradev/delphi/file/DelphiFileTest.java
Cirras: Added `testByteOrderMarkShouldBeStripped`

delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/AbstractDelphiTestFile.java
```diff
@@ -55,6 +55,8 @@ protected InputFile inputFile() {
       file.deleteOnExit();
 
       try (FileWriter fileWriter = new FileWriter(file, UTF_8)) {
+        // Prepend utf-8 BOM
```
zaneduffield, line 58: nit: `\ufeff` is just a codepoint, which isn't inherently UTF-8

fourls: I guess if you want to be technical, it's not even the BOM, just a zero-width non breaking space 😄

In this case, the effect of this line is to write the UTF-8 BOM to the file, so imo the comment's not wrong (though my corresponding nit is that it is wrongly capitalised...)

zaneduffield: Oh no! I've been out-nitted 🥴

fourls, line 58: I think you misunderstood me - my nit was that `utf-8` should be capitalised, especially if `BOM` is 😄

fourls (changes requested):
Looks good, just one note about the comment. Other than that it's good to approve.

delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/AbstractDelphiTestFile.java
Cirras: Oops, I have a bias towards lowercase "utf-8" for some reason.

fourls (approved):
Looks good!

Issue #238, opened 2024-05-17
Bug: PointerName rule wrong issue when type name startswith T without prefix


denieu:
### Prerequisites

- [X] This bug is in SonarDelphi, not SonarQube or my Delphi code.
- [X] This bug has not already been reported.

### SonarDelphi version

1.5.0

### SonarQube version

10.4.1.88267

### Issue description

The PointerName rule checks whether pointers follow the naming convention.
According to the rule "Why is this an issue?" example:
- ^TObject should be named PObject
- ^TMyType should be named PMyType

Using the T prefix for type/class/record names is good practice, but it is something that will not always be followed.
Rules that validate type/class/record naming can have a prefix setting other than T or can be disabled.

Based on the context above, if a type is defined without the T prefix, but the type name starts with T, the PointerName rule report will be invalid.

For example, a pointer called `PTaskInfo` for the custom type `TaskInfo`, which does not have the prefix T, but its name starts with T, will have an invalid issue reported by the PointerName rule.

I believe the problem can be fixed by changing the check done [here](https://github.com/integrated-application-development/sonar-delphi/blob/master/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java#L58), to something like:
```Java
private static String expectedPointerName(String dereferencedName) {
    return "P" + dereferencedName.substring(dereferencedName.startsWith("T") && 
            dereferencedName.length() > 1 && 
            Character.isUpperCase(dereferencedName.charAt(1)) ? 1 : 0);
}
```

Is my thought valid? Does the proposed fix make sense? In this case, I would like to open a PR with the correction, I'm happy to receive your feedback and suggestions.

### Steps to reproduce

Run Sonar Delphi with the code exemplified in this issue.

### Minimal Delphi code exhibiting the issue

```Pascal
unit MyTestUnit;

interface

type
  TaskInfo = record
    title    : string;
    assignee : string;
  end;
  PTaskInfo = ^TaskInfo;

implementation
begin
end.
```

fourls:
Thanks for raising this issue, @denieu! It's a good point. 

I wonder if we should even raise this issue at all if the originating type is non-compliant with the naming convention. For example, with the type `TaskInfo`, all we know is that the naming convention is broken in some way - we can't assume it is `<no prefix> + TaskInfo` rather than `T + askInfo`, because both equally violate the naming convention.

In my mind there's three possible options we can take in cases where the naming convention is broken like this:

[...]

PR #240, opened 2024-05-21
Enhance PointerName rule analysis and documentation

denieu:
This PR improves the behavior of the PointerName rule for types that do not follow naming conventions. It also allows pointers to types with the prefix E and I, in addition to the already supported T.

Fixes https://github.com/integrated-application-development/sonar-delphi/issues/238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - "Move to implementation section" quick fix for `ImportSpecificity`.
 - "Remove unused import" quick fix for `UnusedImport`.
+- More comprehensive analysis and more detailed documentation for the `PointerName` rule.
 - **API:** `UsesClauseNode::getImports` method.
 - **API:** `InterfaceSectionNode::getUsesClause` method.
 - **API:** `ImplementationSectionNode::getUsesClause` method.
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
@@ -18,6 +18,8 @@
  */
 package au.com.integradev.delphi.checks;
 
+import au.com.integradev.delphi.utils.NameConventionUtils;
+import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.plugins.communitydelphi.api.ast.PointerTypeNode;
 import org.sonar.plugins.communitydelphi.api.ast.TypeDeclarationNode;
@@ -31,6 +33,8 @@ import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 @Rule(key = "PointerName")
 public class PointerNameCheck extends DelphiCheck {
   private static final String MESSAGE = "Rename this type to match the expected naming convention.";
+  private static final String POINTER_PREFIX = "P";
+  private static final List<String> EXTENDED_TYPE_PREFIXES = List.of("T", "E", "I");
 
   @Override
   public DelphiCheckContext visit(TypeDeclarationNode type, DelphiCheckContext context) {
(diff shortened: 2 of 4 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
```diff
@@ -31,6 +34,15 @@
...
   private static final String MESSAGE = "Rename this type to match the expected naming convention.";
+  private static final String POINTER_PREFIX = "P";
+  private static final String EXTENDED_TYPE_PREFIXES = "T,E,I";
+
+  private List<String> extendedTypePrefixesList;
+
+  @Override
+  public void start(DelphiCheckContext context) {
+    extendedTypePrefixesList = Splitter.on(',').trimResults().splitToList(EXTENDED_TYPE_PREFIXES);
+  }
```
fourls, line 45: This can be done more concisely by constructing the list directly as the static final variable with `List.of`:

```suggestion
  private static final List<String> EXTENDED_TYPE_PREFIXES = List.of("T", "E", "I");
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
```diff
@@ -41,20 +53,28 @@ public DelphiCheckContext visit(TypeDeclarationNode type, DelphiCheckContext con
     return super.visit(type, context);
   }
 
-  private static boolean isViolation(TypeDeclarationNode type) {
+  private boolean isViolation(TypeDeclarationNode type) {
     if (type.isPointer()) {
```
fourls, line 57: Now that there's more logic in this method, there's a lot of quite indented code - I wonder if it could be a good idea to change this `type.isPointer()` check to be an early return.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
```diff
@@ -41,20 +53,28 @@ public DelphiCheckContext visit(TypeDeclarationNode type, DelphiCheckContext con
...
+          String expected = POINTER_PREFIX + dereferencedName.substring(1);
+          return !typeName.equals(expected);
+        }
+
+        if (NameConventionUtils.compliesWithPascalCase(dereferencedName)) {
+          String expected = POINTER_PREFIX + dereferencedName;
+          return !typeName.equals(expected);
+        }
+
+        return !NameConventionUtils.compliesWithPrefix(typeName, POINTER_PREFIX);
```
fourls, line 75: This currently raises an issue when custom prefixes are used, e.g.
```delphi
type PxyMyClass = ^TxyMyClass;
```

delphi-checks/src/test/java/au/com/integradev/delphi/checks/PointerNameCheckTest.java
```diff
@@ -31,6 +31,11 @@ void testCompliantNameShouldNotAddIssue() {
         .onFile(
             new DelphiTestUnitBuilder()
                 .appendDecl("type")
+                .appendDecl("  PCompliesWithPrefixT = ^TCompliesWithPrefixT;")
+                .appendDecl("  PCompliesWithPrefixE = ^ECompliesWithPrefixE;")
+                .appendDecl("  PCompliesWithPrefixI = ^ICompliesWithPrefixI;")
+                .appendDecl("  PCompliesWithPascalCase = ^CompliesWithPascalCase;")
+                .appendDecl("  PComplientPointer = ^noncomplient_type;")
```
fourls, line 38: Little typo here.
```suggestion
                .appendDecl("  PCompliantPointer = ^noncompliant_type;")
```

delphi-checks/src/test/java/au/com/integradev/delphi/checks/PointerNameCheckTest.java
fourls, line 1: Thanks for updating the check tests for this! It could also be good to have some new tests around custom prefixes, e.g.

Compliant names with custom prefixes:

```delphi
type
  PxyMyClass = ^TxyMyClass;
```

Prefix mismatches:

```delphi
type
  PMyClass = ^TxyMyClass; // Noncompliant
  PxyMyClass = ^TMyClass; // Noncompliant
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/PointerNameCheck.java
fourls, line 1: Since the rule behaviour is changing, the PointerName rule description in [PointerName.html](https://github.com/integrated-application-development/sonar-delphi/blob/master/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/PointerName.html) should also be updated.

fourls (changes requested):
Hi @denieu, thanks for raising this PR!

Overall it looks great - tests pass, formatted, well set up, etc. I've added a couple of notes but they are mostly small things.

The main bug I've noticed is that custom prefixes aren't supported at all - this case we discussed gets erroneously raised as an issue:

```delphi
type PxyMyClass = ^TxyMyClass;
```

I still think that the logic we need can be summarised as:

A pointer type is the result of either

1. Removing the initial `T` / `E` / `I`
2. Adding a `P`

or

1. Adding a `P`

I think this catches all cases except `PxyMyClass = TMyClass`, but there's a fair argument to be made that that should raise an issue anyway, so I'm not particularly bothered about it.

Looking forward to seeing the next iteration 😄

fourls (changes requested):
Thanks @denieu, the code change looks all good now.

Thanks for updating the rule description as well - it looks good overall, and I definitely agree that it should be detailed. That being said, it is probably a little bit long in its current state - I think that once it's trimmed a little it will be good to go.

I've left some notes about possible ways to make it more concise, but feel free to take different approaches if you have any other ideas!

denieu:
Thanks for the review @fourls.

I made the suggested adjustments and also added a description of the changes made to CHANGELOG.md.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/PointerName.html
```diff
@@ -2,12 +2,37 @@ <h2>Why is this an issue?</h2>
 <p>
   Code that follows a consistent naming convention is self-documenting. In Delphi, all pointer names
   should be named identically to the dereferenced type, with the type prefix swapped for
-  <code>P</code>.
+  <code>P</code>, case original prefix are <code>T</code>, <code>E</code> or <code>I</code>.
```
fourls, line 5: The end of this sentence doesn't quite make sense - it's probably okay to remove, since this information is described later on.

```suggestion
  <code>P</code>.
```
Rewording is also OK if you would prefer.

CHANGELOG.md
```diff
@@ -22,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parsing errors on parenthesised anonymous methods.
 - Incorrect ordering of edits in the "Separate grouped parameters" quick fix for `GroupedParameterDeclaration`.
 - SonarLint registration of rules that don't support execution in a SonarLint context.
+- Incorrect handling of some type naming in the `PointerName` rule.
```
fourls, line 26: Thanks for updating the changelog! Usually a maintainer would do this before merging (to avoid merge conflicts).

I'd consider this an improvement rather than a fix, as it's making the rule behaviour meaningfully better rather than fixing a quirk. Because of this, I think it could be summed up in a single bullet point under the `Changed` heading instead of two separate points in different headings.

fourls (changes requested):
Thanks @denieu, almost good to go now. Just a couple wording notes and that's it :smile:

denieu:
@fourls I made the suggested adjustments and also changed the title of the commit and the PR to comply with the changes.

fourls (approved):
Looks good 🎉 Thanks for your effort in submitting this PR, it's a great quality of life improvement for PointerName.

The next official release will be in a week or two, so there's not long to wait until it's in everyone's hands!

PR #246, opened 2024-05-31
Remove warning about Delphi Community Edition being unsupported

Cirras:
Good news!
It turns out that current versions of Delphi CE actually do include the standard library source code.


fourls (approved):
Hooray!

PR #248, opened 2024-06-05
Improve analysis and modeling around property specifiers

Cirras:
The original goal of this PR was just to exclude properties with an `implements` specifier in the `UnusedProperty` rule.

It ended up doing a bit more than that:
* Modeled all of the non-trivial property specifiers as proper AST nodes (`default`, `implements`, `index`, `stored`)
* Fixed a bug where name resolution did not occur for type references in `implements` specifiers
* Excluded properties with an `implements` specifier in `UnusedProperty` (easiest part...)
* Fixed overly-permissive parsing rules where `string` and `file` were erroneously permitted **anywhere** that `typeReference` was allowed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve AST modeling around property specifiers.
 - Improve semantic analysis around property specifiers.
+- Exclude properties with `implements` specifiers in `UnusedProperty`.
 
 ## [1.7.0] - 2024-07-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **API:** `PropertyNameDeclaration::getImplementedTypes` method.
+- **API:** `PropertyNode::getDefaultSpecifier` method.
+- **API:** `PropertyNode::getImplementsSpecifier` method.
+- **API:** `PropertyNode::getIndexSpecifier` method.
+- **API:** `PropertyNode::getStoredSpecifier` method.
+- **API:** `PropertyDefaultSpecifierNode` node type.
+- **API:** `PropertyImplementsSpecifierNode` node type.
+- **API:** `PropertyIndexSpecifierNode` node type.
+- **API:** `PropertyStoredSpecifierNode` node type.
+
+### Changed
+
+- Improve AST modeling around property specifiers.
+- Improve semantic analysis around property specifiers.
+
 ## [1.7.0] - 2024-07-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve semantic analysis around property specifiers.
 - Exclude properties with `implements` specifiers in `UnusedProperty`.
 
+### Fixed
+
+- Overly permissive parsing rules around `string` and `file` types.
+
 ## [1.7.0] - 2024-07-02
 
 ### Added
--- a/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
+++ b/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
@@ -650,16 +650,26 @@ property                     : attributeList? CLASS? PROPERTY nameDeclaration pr
                              ;
 propertyArray                : lbrack! formalParameterList rbrack!
                              ;
-propertyDirective            : ';' DEFAULT
-                             | DEFAULT expression
+propertyDirective            : ';' propertyDefaultNoExpression
+                             | propertyDefault
                              | propertyReadWrite
                              | propertyDispInterface
-                             | IMPLEMENTS typeReference (',' typeReference)*
-                             | INDEX expression
+                             | propertyImplements
+                             | propertyIndex
+                             | propertyStored
                              | NODEFAULT
-                             | STORED expression
+                             ;
+propertyDefaultNoExpression  : DEFAULT<PropertyDefaultSpecifierNodeImpl>^
+                             ;
+propertyDefault              : DEFAULT<PropertyDefaultSpecifierNodeImpl>^ expression
                              ;
 propertyReadWrite            : (READ<PropertyReadSpecifierNodeImpl>^ | WRITE<PropertyWriteSpecifierNodeImpl>^) primaryExpression
+                             ;
+propertyImplements           : IMPLEMENTS<PropertyImplementsSpecifierNodeImpl>^ typeReference (',' typeReference)*
+                             ;
+propertyIndex                : INDEX<PropertyIndexSpecifierNodeImpl>^ expression
+                             ;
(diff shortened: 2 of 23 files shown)


delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
fourls, line 1: Should `weakAliasType` be updated to use `typeReferenceOrStringOrFile`? This would require removing `stringType` and `fileType` from `typeDecl`, but I think it would be more accurate.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -456,9 +457,14 @@ enumType                     : '('<EnumTypeNodeImpl>^ (enumTypeElement (',')?)*
...
 enumTypeElement              : nameDeclaration ('=' expression)? -> ^(TkEnumElement<EnumElementNodeImpl> nameDeclaration expression?)
                              ;
-typeReference                : stringType
+typeReference                : nameReference -> ^(TkTypeReference<TypeReferenceNodeImpl> nameReference)
+                             ;
+typeReferenceOrString        : stringType
+                             | typeReference
+                             ;
+typeReferenceOrStringOrFile  : stringType
                              | FILE<FileTypeNodeImpl>^
```
fourls, line 466: Why is this specifically *just* the `file` keyword and not `fileType`?

fourls (changes requested):
Looks good for the most part - you rugpulled me with the force push 😄 

Just a few questions.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
Cirras: `string[x]` and `file of x` are definitely not weak alias types, in the same way `array of x` is not an alias type.
I think we could **probably** make the argument that standalone `string` and `file` are alias types, though. It feels like an awkward one to model.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
Cirras: `file of x` can't be substituted in place of a name reference, it's only allowed as a type declaration.

Cirras: This probably applies to `string[x]` too, so we may need to adjust the `stringType` usage here.

Cirras: Can confirm that this applies to `string[x]` too, I've replaced these `stringType` usages with `STRING<StringTypeNodeImpl>^`.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -526,17 +526,27 @@ property                     : attributeList? CLASS? PROPERTY nameDeclaration pr
...
-                             | INDEX expression
+                             | propertyImplements
+                             | propertyIndex
+                             | propertyStored
                              | NODEFAULT
-                             | STORED expression
+                             ;
+propertyDefaultNoExpression  : DEFAULT<PropertyDefaultSpecifierNodeImpl>^
+                             ;
+propertyDefault              : DEFAULT<PropertyDefaultSpecifierNodeImpl>^ expression
```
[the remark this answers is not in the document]
fourls: I didn't see the semicolon only applied to `propertyDefaultNoExpression`. All good.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
fourls: I'm not sure I understand - isn't `weakAliasType` the rule for parsing the RHS of a weak alias type definition? And you can certainly do this:

```delphi
type
  TFoos = array of TFoo;
  TString20 = string[20];
  TFileOfFoo = file of TFoo;
```

fourls: Makes sense - I misunderstood the meaning of the `weakAliasType` rule.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
fourls: Gotcha.

CHANGELOG.md
```diff
@@ -10,6 +10,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+
+- Improve AST modeling around property specifiers.
+- Improve semantic analysis around property specifiers.
+- Exclude properties with `implements` specifiers in `UnusedProperty`.
+
+### Fixed
+
+- Overly-permissive parsing rules around `string` and `file` types.
```
fourls, line 31:
```suggestion
- Overly permissive parsing rules around `string` and `file` types.
```

fourls (changes requested):

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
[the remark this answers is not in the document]
Cirras: I ended up doing this:
* added `fullStringType` and `fullFileType` for the full syntax
* stripped `stringType` and `fileType` down to the keyword
* used `fullStringType` and `fullFileType` in `typeDecl` and `varType`

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -545,13 +546,17 @@ arrayIndices                 : lbrack (varType ','?)+ rbrack
...
+fullFileType                 : FILE<FileTypeNodeImpl>^ (OF varType)?
+                             ;
+setType                      : SET<SetTypeNodeImpl>^ OF varType
                              ;
 pointerType                  : '^'<PointerTypeNodeImpl>^ varType
                              ;
-stringType                   : STRING<StringTypeNodeImpl>^ (lbrack! expression rbrack!)?
+stringType                   : STRING<StringTypeNodeImpl>^
+                             ;
+fullStringType               : STRING<StringTypeNodeImpl>^ (lbrack! expression rbrack!)?
```
fourls, line 559: Do you think it's worth removing the discard operators from `lbrack` and `rbrack` here, as part of the ongoing effort to remove these? Might make it tricky to do e.g. a quick fix that changes string types.

Does it also impact the range of the `StringTypeNode` (e.g. make it one character too short on the right, since the last token is removed)?

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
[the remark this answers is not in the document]
fourls: OK, looks good.

fourls (changes requested):
Code looks good - just one thought.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
Cirras: This is a good call, I'll remove the discards.

fourls (approved):
Looks good!

PR #249, opened 2024-06-06
Add extra exclusions for common loop counter names in `ShortIdentifier`

Cirras:
This PR updates the default exclusion list for `ShortIdentifier` to add some common loop counter names (we already had `I`):
* `II`
* `III`
* `J`
* `K`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `RedundantInherited` analysis rule, which flags redundant `inherited` statements that can be safely removed.
 
+### Changed
+
+- Exclude `J` and `K` by default in `ShortIdentifier`.
+
 ## [1.6.0] - 2024-05-31
 
 ### Added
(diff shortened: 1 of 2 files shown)


fourls:
`J` and `K` I accept - do we have any examples of `II` and `III` being accepted as standard loop counter names? I did a quick google and didn't find much. I'm also inclined towards them being actively bad practice, considering how difficult it is to tell them apart.

Cirras:
Alright, you've convinced me to leave out `II`/`III`.

`J` and `K` are the clear winners in terms of popularity, and in the unlikely event that someone is married to their multiple-`I` convention, they can take advantage of the rule's configurability.

(As for me, I'll probably just change my occasional `II` to `J` 😄)

fourls (dismissed):
Yeah, agreed that it makes sense to stay with the mainstream for the defaults - as you say, the rule's configurable. 

Looks good!

fourls (changes requested):
The changelog needs a rebase to resolve merge conflicts.

fourls (approved):

PR #252, opened 2024-06-07
Fix sonar issues in `RedundantInheritedCheck`

jgardn3r:
This change:
- Removes unnecessary parentheses
- Replaces a lambda with a method reference
- Changed the type of an `int` expression to `long`


Cirras (approved):
Looks good to me. 👍

PR #255, opened 2024-06-11
Add found and expected value to `LowercaseKeywordCheck` message

denieu:
A small improvement to the `LowercaseKeywordCheck` rule, now the issue message shows the current value and the expected value, making it easier for users to understand what needs to be fixed.

In addition to making the problem even more intuitive, this change makes it possible to know what the problem is directly from the issues list.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/LowercaseKeywordCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/LowercaseKeywordCheck.java
@@ -34,8 +34,6 @@ import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 @DeprecatedRuleKey(ruleKey = "LowerCaseReservedWordsRule", repositoryKey = "delph")
 @Rule(key = "LowercaseKeyword")
 public class LowercaseKeywordCheck extends DelphiCheck {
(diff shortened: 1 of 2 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/LowercaseKeywordCheck.java
```diff
@@ -34,7 +34,7 @@
 @DeprecatedRuleKey(ruleKey = "LowerCaseReservedWordsRule", repositoryKey = "delph")
 @Rule(key = "LowercaseKeyword")
 public class LowercaseKeywordCheck extends DelphiCheck {
-  private static final String MESSAGE = "Lowercase this keyword.";
+  private static final String MESSAGE = "Lowercase this keyword (found: \"%s\" expected: \"%s\").";
```
Cirras, line 37: Can you inline `MESSAGE` into the usage in `withMessage` to fix the failing build?
ErrorProne doesn't like it when format-strings with 1 usage are pulled out into constants.

Cirras (changes requested):

Cirras (changes requested):

Cirras (approved):
Looks good - thanks! 👍

PR #258, opened 2024-06-13
Treat `result` attribute as case-insensitive in NUnit test reports

Cirras:
This PR fixes a bug where the `result` attribute was treated as a case-sensitive string in NUnit test reports, causing unusual casings to be erroneously reported as `TestResult.Status.FAILURE`.


zaneduffield (approved):
LGTM 👍

PR #264, opened 2024-06-28
Add quick fixes for `LowercaseKeywordCheck`

denieu:
This PR add "Correct to %s" quick fix for `LowercaseKeywordCheck` rule, in a similar way to what already exists for the `MixedNamesCheck` rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- "Correct to (correct case)" quick fix for `LowercaseKeyword`.
 - **API:** `PropertyNameDeclaration::getImplementedTypes` method.
 - **API:** `PropertyNode::getDefaultSpecifier` method.
 - **API:** `PropertyNode::getImplementsSpecifier` method.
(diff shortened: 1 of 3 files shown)


fourls:
Hi @denieu, thanks for this PR - it's a good addition! There's a fair few rules that are good candidates for an easy quick fix, and this one will be very helpful.

Just thought I'd jump in because the quick fix comment syntax is a little unclear, and it's not documented anywhere at the moment. 

This test is failing because you're asserting that there's one quick fix with two text edits, when there's actually two quick fixes with one text edit each. To assert there's two separate quick fixes you need to supply an identifier before the `@` - here's an example from the FormatArgumentCount test:

https://github.com/integrated-application-development/sonar-delphi/blob/58a870ff4923d88c7c9e9ddafb063c80271fda1c/delphi-checks/src/test/java/au/com/integradev/delphi/checks/FormatArgumentCountCheckTest.java#L95-L96

`qf1` and `qf2` are just identifiers I came up with, there's no special meaning to them.

denieu:
Thanks @fourls, saved me a few hours of debugging😁, now it works as expected.

EduardoVaz06 (approved):

fourls (changes requested):
Hi @denieu, code looks good!

Good work making sure that it's on the keyword token, not the entire keyword.
Could we add a test to assert that this is the case? Something like:

```java

  @Test
  void testUppercaseKeywordShouldAddQuickFixOnToken() {
    CheckVerifier.newVerifier()
        .withCheck(createCheck())
        .onFile(
            new DelphiTestUnitBuilder()
                .appendImpl("procedure Foo;")
                .appendImpl("begin")
                .appendImpl("  // Fix@[+1:2 to +1:5] <<for>>")
                .appendImpl("  FOR I := 0 to 5 do // Noncompliant")
                .appendImpl("  begin")
                .appendImpl("  end;")
                .appendImpl("end;"))
        .verifyIssues();
  }
```

Also, just thought I'd mention that PR approvals only count if the reviewer has write access - we will always do a full review before merging a PR, regardless of whether it has been approved by a third party or not.

fourls (dismissed):
Thanks @denieu! Looks great.

fourls (approved):

PR #265, opened 2024-07-01
Fix scan failures on immediately-terminated multiline comments

Cirras:
There's a bug in the lookahead for multiline comments where the first character inside the comment is discarded.
The effect is that an immediately-terminated comment will erroneously consume input until it finds another terminator.

This caused parsing failures in RTL unit `System.Types`.


fourls (approved):
GTM!

Oops, looks like my comment's off by one...

PR #267, opened 2024-07-02
Add support for the library path

### LINKED ISSUE #254: Add support for the library path
The analyzer currently reads only the project search path and ignores the library path.

See #250

Cirras:
This PR adds support for the global library path.

All we had to do was grab the `DelphiLibraryPath` and `DelphiTranslatedLibraryPath` properties and append them to the `DCC_UnitSearchPath` to complete the search path.
These properties are reflected from the registry to `EnvOptions.proj` by the IDE, which is how they become available to MSBuild and by extension the analyzer.

The nice thing is that 95% of the work for this was done a long time ago when we implemented comprehensive support for MSBuild project files.

Closes #254.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Include the global library path in unit import resolution.
 - Improve AST modeling around property specifiers.
 - Improve semantic analysis around property specifiers.
 - Exclude properties with `implements` specifiers in `UnusedProperty`.
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectParser.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectParser.java
@@ -81,11 +81,26 @@ final class DelphiProjectParser {
   }
 
   private List<Path> createSearchDirectories(Path dprojDirectory, ProjectProperties properties) {
-    List<Path> explicitPaths = createPathList(properties, "DCC_UnitSearchPath");
+    /*
+     We manually append the library paths here, even though it's not strictly correct to do so.
 
-    List<Path> allPaths = new ArrayList<>(explicitPaths.size() + 1);
+     CodeGear.Delphi.Targets appends the library paths to DCC_UnitSearchPath to create a new
+     property called UnitSearchPath, which then gets passed through to the compiler.
+
+     However, there are some good reasons not to just read the UnitSearchPath property:
+
+     - It would tie us to an implementation detail of the MSBuild glue in CodeGear.Delphi.Targets.
+     - If the UnitSearchPath property were ever renamed, we'd fall out of compatibility.
+     - Relying on CodeGear.Delphi.Targets details would require us to mock it up in testing.
+    */
+
+    List<Path> allPaths = new ArrayList<>();
+
     allPaths.add(dprojDirectory);
-    allPaths.addAll(explicitPaths);
+    allPaths.addAll(createPathList(properties, "DCC_UnitSearchPath"));
+    allPaths.addAll(createPathList(properties, "DelphiLibraryPath"));
+    allPaths.addAll(createPathList(properties, "DelphiTranslatedLibraryPath"));
+
     return Collections.unmodifiableList(allPaths);
   }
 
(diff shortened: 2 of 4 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectParser.java
```diff
@@ -81,11 +81,32 @@ private static Set<String> createUnitScopeNames(ProjectProperties properties) {
...
+    //
+    // It would tie us to what is arguably an implementation detail of the MSBuild glue in
+    // CodeGear.Delphi.Targets.
+    // If the UnitSearchPath property were ever renamed, we'd have to do something gnarly and
+    // horrible to keep compatibility between Delphi versions. It's less brittle to rely upon the
+    // public DCC_UnitSearchPath property, since that would be quite breaking to rename.
+    //
+    // Additionally, relying on these CodeGear.Delphi.Targets details would require us to mock it up
+    // in testing.
+    //
```
fourls, line 102: This would be better as a single multiline comment IMO.

delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectParser.java
```diff
@@ -81,11 +81,32 @@ private static Set<String> createUnitScopeNames(ProjectProperties properties) {
...
+    // good reasons not to.
+    //
+    // It would tie us to what is arguably an implementation detail of the MSBuild glue in
+    // CodeGear.Delphi.Targets.
+    // If the UnitSearchPath property were ever renamed, we'd have to do something gnarly and
+    // horrible to keep compatibility between Delphi versions. It's less brittle to rely upon the
+    // public DCC_UnitSearchPath property, since that would be quite breaking to rename.
+    //
+    // Additionally, relying on these CodeGear.Delphi.Targets details would require us to mock it up
+    // in testing.
```
fourls, line 101: To make it a bit clearer, these could be made dot points, e.g.

```
There are good reasons not to just read the UnitSearchPath property:

- It would tie us to ....
- If the UnitSearchPath property ...
- Relying on these ...
```

I think they could also be made more concise.

delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectParser.java
```diff
@@ -81,11 +81,32 @@ private static Set<String> createUnitScopeNames(ProjectProperties properties) {
   }
 
   private List<Path> createSearchDirectories(Path dprojDirectory, ProjectProperties properties) {
-    List<Path> explicitPaths = createPathList(properties, "DCC_UnitSearchPath");
+    //
+    // We manually append the library paths here, even though it's not strictly correct to do so.
+    //
+    // CodeGear.Delphi.Targets appends the library paths to DCC_UnitSearchPath to create a new
+    // property called UnitSearchPath.
+    // It's the UnitSearchPath property that then gets passed through to the compiler.
```
fourls, line 89:
```suggestion
    // property called UnitSearchPath, which then gets passed through to the compiler.
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for the global library path.
```
fourls, line 12: It might be good to reword this so it's a little more obvious to the user what this addition means, e.g. "Unit resolution now includes the global library path".

fourls (changes requested):
Code change looks good - just some documentation nitpicks.

Have you tested this on a real project that relies on units in the library path? If yes then I'm happy to approve after the feedback's addressed, otherwise I'll do some testing before approving.

CHANGELOG.md
```diff
@@ -31,6 +32,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.7.0] - 2024-07-02
 
+- The global library path is now included in unit import resolution.
+
+## [1.7.0] - 2024-07-02
+``
```
zaneduffield, line 38: I think this was messed up in the merge conflict.

fourls (commented):
What Zane said.

Apart from that, all looks good to me - I'll do some testing now and then I'll approve if all works well.

CHANGELOG.md
Cirras: Oops! Fixed that.

fourls (approved):
Looks good! I tested by putting a unit on my library path, then importing that unit in a uses clause. Scanning with 1.7.0 produced no issues, but scanning with this PR produced the expected UnusedImport issue, indicating that it has been properly resolved.

PR #270, opened 2024-07-18
Fix bug in CheckVerifier quick fix text edit assertions

fourls:
This PR fixes a bug in the CheckVerifier implementation that causes tests to fail when the number of Sonar-level text replacements does not match the number of SonarDelphi-level quick fix edits. Because one `QuickFixEdit` can correspond to multiple text replacements (e.g. `QuickFixEdit.move`, which replaces two ranges), this test could cause valid quick fixes to fail.

We don't currently have any quick fixes that encounter this problem in tests.


Cirras (approved):
Looks good to me.

PR #276, opened 2024-08-01
Silence noisy warnings around library path properties

Cirras:
In #267, we added support for the library path by manually adding `DelphiLibraryPath` and `DelphiTranslatedLibraryPath` to the search path in `DelphiProjectParser`.

This works well, but produces some noisy warnings when paths in the library path properties don't exist.
By default, the library path contains non-existent paths.

We're already special-casing the construction of the search path here, so what's one more kludge?
This PR silences warnings when constructing path lists for the `DelphiLibraryPath` and `DelphiTranslatedLibraryPath` properties.


fourls (approved):
Looks good! I think this makes sense, invalid paths in the library path are not the "scan's problem" in the same way that invalid paths coming from the dproj are.

I will merge once CI passes.

PR #277, opened 2024-08-01
Fix token index OOB exceptions in `RedundantInherited`

jgardn3r:
When creating quick fixes for `RedundantInherited` the token list is iterated. The index to start at is determined by the token index of the bare `inherited` statement. This index, however, is affected by `{$I}` directives, while the token list is not. This manifested with `ArrayIndexOutOfBoundsException` in cases with an included file.

Prior to traversing the token list, the starting location must be found. Rather than assuming the token index matches 1-to-1 with the token list.


fourls (changes requested):
Looks good other than a function rename.

In the near future we should definitely abstract some of this token traversal stuff into a utility  - with quick fixes we're going to need to be doing a lot more of this kind of processing.

fourls:
Also, confirmed that this has resolved the exceptions caught by our QA process.

fourls (approved):
Looks good TM!

PR #279, opened 2024-08-05
Treat `out` parameters as uninitialized in `VariableInitialization`

### LINKED ISSUE #134: `VariableInitialization` rule should treat `out` parameters as uninitialized
### Prerequisites

### Rule to improve

VariableInitialization

#### Current behavior

Routine parameters are currently always considered initialized at the start of a routine.

#### Suggested behavior

`out` parameters should be considered uninitialized at the start of a routine.

### Rationale

`out` parameters are meant to be routine results, and should never be assumed to be initialized.
Indeed, if they are initialized then it's actually more of a smell (see #133).

Cirras:
This PR changes the `VariableInitialization` rule to treat `out` parameters as uninitialized at the start of a routine.

Closes #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `out` parameters are treated as uninitialized at the start of a routine in
+  `VariableInitialization`.
+
 ### Fixed
 
 - Parsing errors around partial `asm` blocks blocks in conditional branches.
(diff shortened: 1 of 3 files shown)


fourls (changes requested):
Looks good.

Could we add tests for

- a nested routine with an `out` parameter
- a nested routine that uses the `out` parameter of its containing routine

I think nested routines are the sort of edge cases that would be good to catch in a test.

fourls (approved):
LGreatTM!

fourls (changes requested):
Should have expected the changelog merge conflict...

fourls (approved):
👍

PR #280, opened 2024-08-07
Fix FPs on case items with multiple values in `CaseStatementSize`

Cirras:
This PR improves the `CaseStatementSize` rule, counting `case` items with multiple values as if they were flattened out to multiple case items.

For example, the following `case` statement will no longer raise an issue:
```pas
case Foo of
  123, 456: Bar;
end;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **API:** `CaseItemStatementNode::getExpressions` method.
+
 ### Changed
 
 - `out` parameters are treated as uninitialized at the start of a routine in
@@ -15,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Parsing errors around partial `asm` blocks blocks in conditional branches.
+- False positives around case items with multiple values in `CaseStatementSize`.
 
 ## [1.8.0] - 2024-08-02
 
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/CaseStatementSizeCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/CaseStatementSizeCheck.java
@@ -19,7 +19,9 @@
 package au.com.integradev.delphi.checks;
 
 import org.sonar.check.Rule;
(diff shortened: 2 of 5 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/CaseStatementSizeCheck.java
```diff
@@ -28,13 +30,40 @@
...
+            expression -> {
+              if (expression instanceof RangeExpressionNode) {
+                // HACK:
+                //  We're assuming that the range expression actually represents 2 or more values,
+                //  and isn't something silly like 0..0.
+                //  Constant expression evaluation would be needed to get the actual number of
+                //  values in the range.
+                //
+                //  See:
+                //   https://github.com/integrated-application-development/sonar-delphi/issues/116
```
fourls, line 62: I don't think this is a hack -  it seems sensible to exclude case statements with a range expression, since they're not directly mappable onto an if statement.

I don't think the mindset should be that we're hacking the behaviour of this rule until we have constant expression evaluation - I would argue that would be overkill in this case. We also probably want to keep the hack comments isolated to where they're most hacky.

CHANGELOG.md
fourls, line 1: The changelog needs to be rebased.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
fourls, line 9: Include an entry for the new API method `CaseItemStatementNode::getExpressions`.

fourls (changes requested):
Code looks good! A couple of minor things.

CHANGELOG.md
Cirras: Good catch.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/CaseStatementSizeCheck.java
Cirras: I'm of two minds about it because it would be nice to be able to statically recognize stupid patterns like a range of 1.
But on balance, I think you're right - it's overkill to consider this a hack.

Cirras: I've changed the comment.

CHANGELOG.md
Cirras: Done.

fourls (approved):
Looks good! 🚀

PR #281, opened 2024-08-07
Add control flow graph implementation

jgardn3r:
This PR adds the notion of a control flow graph. Control flow graphs are a stepping stone in data flow analysis and symbolic execution.

With just the control flow graph alone, two new rules could be implemented:
* RedundantJump
* LoopExecutingAtMostOnce

In addition to that, some more API traversal methods were added:
* `RepeatStatementNode::getGuardExpression`
* `RepeatStatementNode::getStatements`
* `CaseStatementNode::getSelectorExpression`
* `CaseItemStatementNode::getStatement`

The thinking is that the control flow graph will be internal for a while before being exposed in the stable API. In light of this, I have made a `ControlFlowGraph` interface in the API, but there are no public ways to create/interact with it.

### DIFF
--- a/delphi-checks-testkit/src/test/java/au/com/integradev/delphi/checks/verifier/CheckVerifierImplTest.java
+++ b/delphi-checks-testkit/src/test/java/au/com/integradev/delphi/checks/verifier/CheckVerifierImplTest.java
@@ -21,14 +21,29 @@ package au.com.integradev.delphi.checks.verifier;
 import static org.assertj.core.api.Assertions.*;
 
 import au.com.integradev.delphi.builders.DelphiTestUnitBuilder;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Named;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.sonar.check.Rule;
 import org.sonar.plugins.communitydelphi.api.ast.ArgumentListNode;
+import org.sonar.plugins.communitydelphi.api.ast.BinaryExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiAst;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
+import org.sonar.plugins.communitydelphi.api.ast.IntegerLiteralNode;
 import org.sonar.plugins.communitydelphi.api.ast.NameReferenceNode;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext.Location;
 import org.sonar.plugins.communitydelphi.api.reporting.QuickFix;
 import org.sonar.plugins.communitydelphi.api.reporting.QuickFixEdit;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.TypeNameDeclaration;
@@ -367,6 +382,214 @@ class CheckVerifierImplTest {
     assertThatThrownBy(verifier::verifyIssueOnProject).isInstanceOf(AssertionError.class);
   }
 
+  private static Stream<Arguments> validSecondaryLocationsCases() {
+    return Stream.of(
+        Arguments.of(Named.of("InOrder", "(1) (2) (3)"), "(-2) (-1)"),
+        Arguments.of(Named.of("OutOfOrder", "(3) (1) (2)"), "(-1) (-2)"),
+        Arguments.of(Named.of("ExtraSpaces", "   ( 1  )   (2 ) ( 3)"), "(-2   ) (-1  ) "));
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("validSecondaryLocationsCases")
+  void testValidSecondaryLocationsOnIssue(String annotation1, String annotation2) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseDirectionalSecondariesOnIntegersCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 2; // Noncompliant " + annotation1)
+                        .appendDecl("  Foo1 = 2;")
+                        .appendDecl("  Foo2 = 2;")
+                        .appendDecl("  Foo3 = 2;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Bar1 = 1;")
+                        .appendDecl("  Bar2 = 1; // Noncompliant " + annotation2));
+
+    assertThatCode(verifier.get()::verifyIssues).doesNotThrowAnyException();
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  private static Stream<Arguments> invalidSecondaryLocationsCases() {
+    return Stream.of(
+        Arguments.of(Named.of("MissingLocation", "(1) (2)"), "(-2)"),
+        Arguments.of(Named.of("ExtraLocation", "(1) (2) (3) (0)"), "(-2) (0) (-1)"),
+        Arguments.of(Named.of("ErroneousFlow", "(1, 0) (2) (3)"), "(-2, 1) (-1)"),
+        Arguments.of(Named.of("MissingLocation", "(1) (3)"), "(-2)"));
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("invalidSecondaryLocationsCases")
+  void testInvalidSecondaryLocationsOnIssue(String annotation1, String annotation2) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseDirectionalSecondariesOnIntegersCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 2; // Noncompliant " + annotation1)
+                        .appendDecl("  Foo1 = 2;")
+                        .appendDecl("  Foo2 = 2;")
+                        .appendDecl("  Foo3 = 2;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Bar1 = 1;")
+                        .appendDecl("  Bar2 = 1; // Noncompliant " + annotation2));
+
+    assertThatThrownBy(verifier.get()::verifyIssues)
+        .hasMessageContainingAll("Issues were expected at", "unexpected at")
+        .isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("invalidSecondaryLocationsCases")
+  void testInvalidSecondaryLocationsOnIssueWithOffset(String annotation1, String annotation2) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseDirectionalSecondariesOnIntegersCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 2;")
+                        .appendDecl("  Foo1 = 2; // Noncompliant@-1 " + annotation1)
+                        .appendDecl("  Foo2 = 2;")
+                        .appendDecl("  Foo3 = 2;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Bar1 = 1; // Noncompliant@+1 " + annotation2)
+                        .appendDecl("  Bar2 = 1;"));
+
+    assertThatThrownBy(verifier.get()::verifyIssues)
+        .hasMessageContainingAll("Issues were expected at", "unexpected at")
+        .isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  private static Stream<Arguments> validFlowsLocationsCases() {
+    return Stream.of(
+        Arguments.of(Named.of("InOrder", "(-3, -1, 2, 3) (-2, 1)")),
+        Arguments.of(Named.of("OutOfOrder", "(-2, 1) (-3, -1, 2, 3)")),
+        Arguments.of(Named.of("ExtraSpaces", "  (   -3,  -1    ,2, 3   )    (-2 , 1)")));
+  }
+
+  @ParameterizedTest
+  @MethodSource("validFlowsLocationsCases")
+  void testValidFlowsLocations(String flows) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseFlowOnEachIntegerLiteralForBinaryExpressionCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 0;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Foo1 = 0;")
+                        .appendDecl("  Bar1 = 0 + 1; // Noncompliant " + flows)
+                        .appendDecl("  Bar2 = 1;")
+                        .appendDecl("  Foo2 = 0;")
+                        .appendDecl("  Foo3 = 0;"));
+
+    assertThatCode(verifier.get()::verifyIssues).doesNotThrowAnyException();
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  @ParameterizedTest
+  @MethodSource("validFlowsLocationsCases")
+  void testValidFlowsLocationsWithOffset(String flows) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseFlowOnEachIntegerLiteralForBinaryExpressionCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 0;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Foo1 = 0;")
+                        .appendDecl("  Bar1 = 0 + 1;")
+                        .appendDecl("  Bar2 = 1; // Noncompliant@-1 " + flows)
+                        .appendDecl("  Foo2 = 0;")
+                        .appendDecl("  Foo3 = 0;"));
+
+    assertThatCode(verifier.get()::verifyIssues).doesNotThrowAnyException();
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  private static Stream<Arguments> invalidFlowsLocationsCases() {
+    return Stream.of(
+        Arguments.of(Named.of("LocationsOutOfOrder", "(-1, 2, 3, -3) (1, -2)")),
+        Arguments.of(Named.of("MissingFlow", "(-3, -1, 2, 3)")),
+        Arguments.of(Named.of("ExtraFlow", "(-1, 2, 3, -3) (1, -2) ()")),
+        Arguments.of(Named.of("MissingLocation", "(-3, -1, 2) (-2)")),
+        Arguments.of(Named.of("EmptyFlow", "()")),
+        Arguments.of(Named.of("NoFlows", "")));
+  }
+
+  @ParameterizedTest
+  @MethodSource("invalidFlowsLocationsCases")
+  void testInvalidFlowsLocationsOnIssue(String annotation) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseFlowOnEachIntegerLiteralForBinaryExpressionCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 0;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Foo1 = 0;")
+                        .appendDecl("  Bar1 = 0 + 1; // Noncompliant " + annotation)
+                        .appendDecl("  Bar2 = 1;")
+                        .appendDecl("  Foo2 = 0;")
+                        .appendDecl("  Foo3 = 0;"));
+
+    assertThatThrownBy(verifier.get()::verifyIssues)
+        .hasMessageContainingAll("Issues were expected at", "unexpected at")
+        .isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
+  @ParameterizedTest
+  @MethodSource("invalidFlowsLocationsCases")
+  void testInvalidFlowsLocationsOnIssueWithOffset(String annotation) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(new WillRaiseFlowOnEachIntegerLiteralForBinaryExpressionCheck())
+                .onFile(
+                    new DelphiTestUnitBuilder()
+                        .appendDecl("const")
+                        .appendDecl("  Foo0 = 0;")
+                        .appendDecl("  Bar0 = 1;")
+                        .appendDecl("  Foo1 = 0; // Noncompliant@+1 " + annotation)
+                        .appendDecl("  Bar1 = 0 + 1;")
+                        .appendDecl("  Bar2 = 1;")
+                        .appendDecl("  Foo2 = 0;")
+                        .appendDecl("  Foo3 = 0;"));
+
+    assertThatThrownBy(verifier.get()::verifyIssues)
+        .hasMessageContainingAll("Issues were expected at", "unexpected at")
+        .isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyNoIssues).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnFile).isInstanceOf(AssertionError.class);
+    assertThatThrownBy(verifier.get()::verifyIssueOnProject).isInstanceOf(AssertionError.class);
+  }
+
   @Test
   void testFileIssue() {
     CheckVerifier verifier =
@@ -492,6 +715,106 @@ class CheckVerifierImplTest {
         reportIssue(context, nameReference, MESSAGE);
       }
       return super.visit(nameReference, context);
+    }
+  }
+
+  /**
+   * This check will add secondary locations on: the same even integer after the first instance; and
+   * the same odd integer before the last instance.
+   */
+  @Rule(key = "WillRaiseDirectionalSecondariesOnIntegersCheck")
+  public static class WillRaiseDirectionalSecondariesOnIntegersCheck extends DelphiCheck {
+
+    private final Map<Integer, List<IntegerLiteralNode>> locations = new HashMap<>();
+
+    @Override
+    public DelphiCheckContext visit(DelphiAst ast, DelphiCheckContext context) {
+      context = super.visit(ast, context);
+      for (Entry<Integer, List<IntegerLiteralNode>> entry : locations.entrySet()) {
+        Integer key = entry.getKey();
+        List<IntegerLiteralNode> literals = entry.getValue();
+        IntegerLiteralNode node;
+        if (key % 2 == 0) {
+          node = literals.remove(0);
+        } else {
+          node = literals.remove(literals.size() - 1);
+        }
+
+        context
+            .newIssue()
+            .onNode(node)
+            .withMessage(MESSAGE)
+            .withSecondaries(
+                literals.stream()
+                    .map(literal -> new Location(MESSAGE, literal))
+                    .collect(Collectors.toList()))
+            .report();
+      }
+      return context;
+    }
+
+    @Override
+    public DelphiCheckContext visit(
+        IntegerLiteralNode integerLiteralNode, DelphiCheckContext context) {
+      Integer number = integerLiteralNode.getValue().intValue();
+      if (this.locations.containsKey(number)) {
+        locations.get(number).add(integerLiteralNode);
+      } else {
+        List<IntegerLiteralNode> numberLocations = new ArrayList<>();
+        numberLocations.add(integerLiteralNode);
+        locations.put(number, numberLocations);
+      }
+
+      return context;
+    }
+  }
+
+  @Rule(key = "WillRaiseFlowOnEachIntegerLiteralForBinaryExpression")
+  public static class WillRaiseFlowOnEachIntegerLiteralForBinaryExpressionCheck
+      extends DelphiCheck {
+
+    private final List<BinaryExpressionNode> binaryExpressions = new ArrayList<>();
+    private final Map<Integer, List<Location>> locations = new HashMap<>();
+
+    @Override
+    public DelphiCheckContext visit(DelphiAst ast, DelphiCheckContext context) {
+      context = super.visit(ast, context);
+      for (BinaryExpressionNode node : binaryExpressions) {
+        IntegerLiteralNode left = node.getLeft().getFirstDescendantOfType(IntegerLiteralNode.class);
+        IntegerLiteralNode right =
+            node.getRight().getFirstDescendantOfType(IntegerLiteralNode.class);
+        if (left == null || right == null) {
+          continue;
+        }
+
+        List<Location> leftLocations = this.locations.get(left.getValue().intValue());
+        List<Location> rightLocations = this.locations.get(right.getValue().intValue());
+        context
+            .newIssue()
+            .onNode(node)
+            .withMessage(MESSAGE)
+            .withFlows(List.of(leftLocations, rightLocations))
+            .report();
+      }
+      return context;
+    }
+
+    @Override
+    public DelphiCheckContext visit(
+        IntegerLiteralNode integerLiteralNode, DelphiCheckContext context) {
+      Integer number = integerLiteralNode.getValue().intValue();
+      if (!locations.containsKey(number)) {
+        this.locations.put(number, new ArrayList<>());
+      }
+      this.locations.get(number).add(new Location(MESSAGE, integerLiteralNode));
+      return context;
+    }
+
+    @Override
+    public DelphiCheckContext visit(
+        BinaryExpressionNode binaryExpressionNode, DelphiCheckContext context) {
+      this.binaryExpressions.add(binaryExpressionNode);
+      return context;
     }
   }
 
--- a/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/LoopExecutingAtMostOnce.html
+++ b/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/LoopExecutingAtMostOnce.html
@@ -1,0 +1,63 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Loops with at most one iteration are equivalent to an <code>if</code> statement. Using loops in
+  this case makes the code less readable.
+</p>
+<p>
+  If the intention was to execute the loop once, an <code>if</code> statement may be used or the
+  loop removed.
+  Otherwise, the jumping statement should be made conditional so the loop can execute more than
+  once.
+</p>
+<p>
+  Loops with at most one iteration can happen with a statement that unconditionally transfers
+  control is misplaced inside the body of the loop.
+</p>
+<p>
+  These statements are:
+</p>
+<ul>
+  <li><code>Exit</code></li>
+  <li><code>Break</code></li>
+  <li><code>Halt</code></li>
+  <li><code>raise</code></li>
+  <li><code>goto</code></li>
+</ul>
+
+<h2>How to fix it</h2>
+<p>
+  Make the statement that affects execution of the loop conditional, or remove it all together.
+</p>
+<pre data-diff-id="1" data-diff-type="noncompliant">
+var I := 0;
+while I &lt; 10 do begin
+  Inc(I);
+  Break; // Noncompliant
+end;
+</pre>
+<pre data-diff-id="2" data-diff-type="noncompliant">
+for var I := 0 to 10 do begin
+  if I = 2 then
+    Break // Noncompliant
+  else begin
+    Writeln(I);
+    Exit; // Noncompliant
+  end;
+end;
+</pre>
+<h4>Compliant solution</h4>
+<pre data-diff-id="1" data-diff-type="compliant">
+var I := 0;
+while I &lt; 10 do begin
+  Inc(I);
+end;
+</pre>
+<pre data-diff-id="2" data-diff-type="compliant">
+for var I := 0 to 10 do begin
+  if I = 2 then
+    Break
+  else begin
+    Writeln(I);
+  end;
+end;
+</pre>
--- a/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/RedundantJump.html
+++ b/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/RedundantJump.html
@@ -1,0 +1,26 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Unnecessarily annotating the default control flow of a program using statements such as
+  <code>Continue</code>, <code>Break</code>, and <code>Exit</code> makes the code harder to read and
+  understand. Although these statements appear to alter the flow of a program, the control flow is
+  identical without it.
+</p>
+<h2>How to fix it</h2>
+<p>Remove the redundant jump:</p>
+<pre data-diff-id="1" data-diff-type="noncompliant">
+procedure Example;
+begin
+  while Condition do begin
+    // ...
+    Continue; // Noncompliant
+  end;
+end;
+</pre>
+<pre data-diff-id="1" data-diff-type="compliant">
+procedure Example;
+begin
+  while Condition do begin
+    // ...
+  end;
+end;
+</pre>
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/CheckTestNameTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/CheckTestNameTest.java
@@ -28,6 +28,7 @@ import com.tngtech.archunit.core.domain.JavaClasses;
 import com.tngtech.archunit.core.domain.JavaMember;
 import com.tngtech.archunit.core.domain.JavaMethod;
 import com.tngtech.archunit.core.domain.JavaModifier;
+import com.tngtech.archunit.core.domain.properties.CanBeAnnotated.Predicates;
 import com.tngtech.archunit.core.domain.properties.HasName;
 import com.tngtech.archunit.core.importer.ClassFileImporter;
 import com.tngtech.archunit.lang.ArchCondition;
@@ -35,6 +36,7 @@ import com.tngtech.archunit.lang.ConditionEvents;
 import com.tngtech.archunit.lang.SimpleConditionEvent;
 import java.util.List;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
 
 class CheckTestNameTest {
@@ -123,6 +125,10 @@ class CheckTestNameTest {
     methods()
         .that(VERIFY_ISSUES)
         .and(not(TESTING_IMPLEMENTATION_DETAILS))
+        .and(
+            DescribedPredicate.or(
+                Predicates.annotatedWith(Test.class),
+                Predicates.annotatedWith(ParameterizedTest.class)))
         .should()
         .haveNameMatching(".*ShouldAdd(Issues?|QuickFix(es)?)$")
         .allowEmptyShould(true)
@@ -135,6 +141,10 @@ class CheckTestNameTest {
         .that(VERIFY_NO_ISSUES)
         .and(not(TESTING_IMPLEMENTATION_DETAILS))
         .and(not(CALL_ASSERT_THROW_BY))
+        .and(
+            DescribedPredicate.or(
+                Predicates.annotatedWith(Test.class),
+                Predicates.annotatedWith(ParameterizedTest.class)))
         .should()
         .haveNameMatching(".*ShouldNotAddIssues?$")
         .allowEmptyShould(true)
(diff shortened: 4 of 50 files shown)


fourls (changes requested):
I've gone through all the core CFG stuff now, I've got a few initial comments that I'll put out so we can get started. I have more reviewing to do still, don't worry 😄 

Overall, looks very cool and seems to work well! I have a few broad thoughts/questions:

- The core CFG stuff relies heavily on public fields, which is very uncomfortably un-encapsulated in my view. I found it difficult to follow since there was very little "contract" about what could change. I see this comes directly from SonarJava - can we deviate from them and redesign for more encapsulation and better OO vibes?

- In something this logically complex I think it'll be vital to have plenty of explanatory comments if we're going to have any chance of maintenance + continual improvement. At the moment it's very sparse. 

- Rules shouldn't instantiate `...Impl` classes themselves, that should be done through the API layer (e.g. a `ControlFlowGraphFactory` that builds a new `ControlFlowGraph`, for example). Core rules can use `Impl` behaviour by casting to `ControlFlowGraphImpl` if they absolutely need to.

- I don't really see the need to keep the `ControlFlowGraph` interface empty for the time being, although if this is something that you think would be beneficial then I don't really mind. Would the API change all that much after this PR is merged?

jgardn3r:
I didn't respond to each review comment because many things have changed and it is practically a rewrite. I have tried to:
1. Improve documentation and explanatory comments
2. Use immutability
3. Create an API that isn't public

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/ExitPath.java
```diff
@@ -0,0 +1,38 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public interface ExitPath extends Successors {
+  /** Next block without exceptional circumstances */
```
Cirras, line 25: Across the board, I think we need to tidy the javadocs in the API layer.
- aim for less terse wording
- use `@param` and `@return` tags
- put javadocs on the types themselves and not just the methods

Cirras (changes requested):

jgardn3r:
There are a few aspects of the current shape that I am not completely happy with yet that I think would be worth discussing/brainstorming:
* The modelling of the _successors_
  * `getSuccessors()`
    * Returns an interface that allows you to check what kind of successors there are
    * Is this worth it?
    * This replaces the `sonar-java` approach of having a flag for a bunch of different things, e.g., `isFinallyBlock`, `isCatchBlock`
  * `getSuccessorBlocks()`
    * This is the successor-type agnostic way of getting all of the successor blocks
    * Generically useful for traversing blocks in rules, etc.
  * Object reference permanence - not necessarily a bad thing, just worth noting
    * When constructing the _blocks_ the successors are treated as immutable
    * If there is a change to the successors a new successor object is used to replace
    * Blocks reference each other using address, rather than ID
  * An idea I had is only making the strongly coupled object as the end and using IDs internally

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/BlockImpl.java
```diff
@@ -0,0 +1,83 @@
...
+  public Set<Block> getSuccessorBlocks() {
+    return Collections.unmodifiableSet(successors.getSuccessorBlocks());
+  }
+
+  @Override
+  public Set<Block> getPredecessorBlocks() {
+    return Collections.unmodifiableSet(predecessors);
+  }
+
+  /** This is allows the mutation `predecessors` */
```
Cirras, line 75: This comment can be removed, it's not adding much.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnknownException.java
```diff
@@ -0,0 +1,37 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public interface UnknownException extends Successors {
```
Cirras, line 25: Can you explain the name of this `Successors` type?
I don't really get it at a glance (javadoc might help).

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
```diff
@@ -0,0 +1,40 @@
...
+  int getId();
+
+  /** Successor blocks of the block */
+  Set<Block> getSuccessorBlocks();
+
+  /** Predecessors of the block */
+  Set<Block> getPredecessorBlocks();
+
+  /** Successors handler of the block */
+  Successors getSuccessors();
```
Cirras, line 36: I'm feeling a bit weird about:
* `getSuccessorBlocks` and `getSuccessors` coexisting at this level of the API
* the asymmetry due to no neatly-typed `getPredecessors` matching up with `getSuccessors`

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.List;
+import java.util.Set;
+import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
+
+public interface Block {
+  /** Unique block identifier */
+  int getId();
```
Cirras, line 27: Something we've pretty directly ripped off from sonar-java.
The unique identifier is all well and good - but what's it actually for?

The only usage in application code seems to be ordering a `PriorityQueue` based on them (?) which I'd have questions about. Are we relying on the implementation detail of a linear ordering based on creation order for this "unique block identifier", or am I misunderstanding something?

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphImpl.java
```diff
@@ -0,0 +1,320 @@
...
+import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
+import org.sonar.plugins.communitydelphi.api.ast.GotoStatementNode;
+import org.sonar.plugins.communitydelphi.api.ast.LabelStatementNode;
+import org.sonar.plugins.communitydelphi.api.ast.NameReferenceNode;
+import org.sonar.plugins.communitydelphi.api.type.Type;
+import org.sonarsource.analyzer.commons.collections.ListUtils;
+
+public class ControlFlowGraphImpl implements ControlFlowGraph {
+
+  private Block currentBlock;
```
Cirras, line 52:
```suggestion
public class ControlFlowGraphImpl implements ControlFlowGraph {
  private Block currentBlock;
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
+import java.util.List;
+import org.sonar.plugins.communitydelphi.api.ast.AnonymousMethodNode;
+import org.sonar.plugins.communitydelphi.api.ast.CompoundStatementNode;
+import org.sonar.plugins.communitydelphi.api.ast.RoutineImplementationNode;
+import org.sonar.plugins.communitydelphi.api.ast.StatementListNode;
+import org.sonar.plugins.communitydelphi.api.ast.StatementNode;
+import org.sonarsource.analyzer.commons.collections.ListUtils;
+
+public class ControlFlowGraphFactory {
```
Cirras, line 30: Make this utility class `final`.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
Cirras: You commented on this above, so I'm just getting some initial thoughts out about the current API shape.

At minimum, maybe you can speak to the reason for the asymmetry here?
`Successors` got its own `Successors` type hierarchy, while predecessors are just the basic `Set<Block>`.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphFactory.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public static ControlFlowGraph create(StatementListNode statements) {
+    return create(statements.getStatements());
+  }
+
+  public static ControlFlowGraph create(List<StatementNode> statements) {
+    ControlFlowGraphImpl cfg = new ControlFlowGraphImpl();
+    ControlFlowGraphVisitor visitor = new ControlFlowGraphVisitor();
+    ListUtils.reverse(statements).forEach(statement -> statement.accept(visitor, cfg));
```
Cirras, line 54: Use Guava's `Lists.reverse` instead.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphImpl.java
```diff
@@ -0,0 +1,320 @@
...
+        .collect(Collectors.toSet());
+  }
+
+  public void popTryContext() {
+    tryContexts.pop();
+  }
+
+  @Override
+  public List<Block> getBlocks() {
+    return Collections.unmodifiableList(ListUtils.reverse(blocks));
```
Cirras, line 277: Use Guava's `Lists.reverse` instead.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/BlockImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    return this.id;
+  }
+
+  public void setId(int id) {
+    this.id = id;
+  }
+
+  @Override
+  public List<DelphiNode> getElements() {
+    return Collections.unmodifiableList(ListUtils.reverse(elements));
```
Cirras, line 53: Use Guava's `Lists.reverse` instead.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
jgardn3r: Thinking about it a little more, I think the `Successors` API is an indirect way of being able to check what type of block a block is. All the while maintaining the object permanence as mentioned.
The `getSuccessorBlocks` endpoint is the opaque way of accessing the successors and is what would be primarily utilised to traverse the graph. The `getSuccessors` on the other hand is a more strongly typed idea, indicated what kinds of successors there are.

Reasoning from my use of it here, and the usages of it in `sonar-java`, the predecessors are just a basic set because the graph is directed. A node only really knows about where it goes next, the fact the predecessors are there at all is a matter of convenience, I would suggest.

I think a solution that makes a lot of sense would be one that incorporates a few things I mentioned:
* Move the type hierarchy to the `Block` type
* Use IDs in the construction, and only construct the object referenced links after pruning
* Alter the API back to the original `sonar-java` approach of just `getSuccessors` and `getPredecessors` on the main `Block`

Let me know what you think.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnknownException.java
jgardn3r: I will add a javadoc.
High level though, if a block can raise an exception but the type of which is unknown, that is when this block occurs.
It just means that there are multiple potential successors, rather than a successor specifically directed to an arm of a try-catch-finally.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
Cirras: Yeah, I'm on board with this idea.
It takes the good part of the SonarJava API while also avoiding boolean flag hell on `Block`.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Block.java
jgardn3r: After discussion, the notion of the ID will remain. However, it will not be in the API and predominantly used for debugging.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
```diff
@@ -226,20 +224,20 @@ private Consumer<DelphiNode> unaryOpTest(UnaryOperator operator) {
...
     checker(block(element(NameReferenceNode.class, "Foo")).succeedsTo(0)).check(cfg);
     Block entry = cfg.getEntryBlock();
-    assertThat(entry.getSuccessorBlocks()).as("1st block is not an exit").isNotEmpty();
-    assertThat(entry.getSuccessorBlocks()).as("number of succeedsTo").hasSize(1);
-    Block exit = entry.getSuccessorBlocks().iterator().next();
-    assertThat(exit.getSuccessorBlocks()).as("2nd block is an exit").isEmpty();
+    assertThat(entry.getSuccessors()).as("1st block is not an exit").isNotEmpty();
+    assertThat(entry.getSuccessors()).as("number of succeedsTo").hasSize(1);
+    Block exit = entry.getSuccessors().iterator().next();
+    assertThat(exit.getSuccessors()).as("2nd block is an exit").isEmpty();
```
Cirras, line 240: For some of these, consider using `withFailMessage` instead of `as`.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/checker/BlockChecker.java
```diff
@@ -71,24 +71,21 @@ private BlockChecker(ElementChecker... elementCheckers) {
...
     for (int elementId = 0; elementId < elementCheckers.size(); elementId++) {
-      elementCheckers
-          .get(elementId)
-          .withBlockId(block.getId(), elementId)
-          .check(block.getElements().get(elementId));
+      elementCheckers.get(elementId).check(block.getElements().get(elementId));
     }
     assertThat(successorChecker)
-        .as(getBlockId(block) + " should have its successors declared")
+        .as(getBlockDisplay(block) + " should have its successors declared")
```
Cirras, line 80: Consider using `withFailMessage` instead of `as`.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/checker/GraphChecker.java
```diff
@@ -0,0 +1,65 @@
...
+    assertThat(exitBlock.getElements()).isEmpty();
+    assertThat(exitBlock.getSuccessors()).isEmpty();
+    assertThat(cfg.getBlocks())
+        .as("CFG entry block is no longer in the list of blocks!")
+        .contains(cfg.getEntryBlock());
+  }
+
+  private void checkLinkedBlocks(String type, List<Block> blocks, Set<Block> linkedBlocks) {
+    for (Block block : linkedBlocks) {
+      assertThat(block).as(type + " block " + " is missing from he list of blocks").isIn(blocks);
```
Cirras, line 62:
```suggestion
      assertThat(block).as(type + " block " + " is missing from the list of blocks").isIn(blocks);
```

Cirras (changes requested):

delphi-frontend/src/main/java/org/sonar/plugins/communitydelphi/api/ast/CaseItemStatementNode.java
```diff
@@ -22,4 +22,6 @@
 
 public interface CaseItemStatementNode extends StatementNode {
   List<ExpressionNode> getExpressions();
+
+  StatementNode getStatement();
```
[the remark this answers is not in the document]
jgardn3r: Yes it can, I will add `@Nullable` to it.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/RedundantJump.html
```diff
@@ -0,0 +1,25 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Unnecessarily using control flow intrinsics makes the code harder to read and understand with no
+  benefit. These statements appear to jump in the code, however, the jump is to the location that
+  would be next anyway.
+</p>
+<h2>How to fix it</h2>
+<p>Remove the redundant jump:</p>
+<pre data-diff-id="1" data-diff-type="noncompliant">
+procedure Example(Foo: TFoo);
```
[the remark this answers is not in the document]
jgardn3r: Probably because I copied it from somewhere else. Removed.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/CaseItemStatementNodeImpl.java
```diff
@@ -43,4 +44,9 @@ public <T> T accept(DelphiParserVisitor<T> visitor, T data) {
   public List<ExpressionNode> getExpressions() {
     return findChildrenOfType(ExpressionNode.class);
   }
+
+  @Override
+  public StatementNode getStatement() {
+    return getFirstChildOfType(StatementNode.class);
+  }
```
Cirras, line 51: Add `@Nullable` to match the API.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
```diff
@@ -0,0 +1,1247 @@
...
+    return buildCFG(Collections.emptyMap(), input);
+  }
+
+  private ControlFlowGraph buildCFG(List<String> variables, String input) {
+    return buildCFG(Map.of("var", variables), input);
+  }
+
+  private ControlFlowGraph buildCFG(Map<String, List<String>> sections, String input) {
+    try {
+      var tempFile = File.createTempFile("CFGTest-", ".pas");
```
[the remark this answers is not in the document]
Cirras: Cool, but useless suggestion in the context of test code.

Cirras (changes requested):
There are some incomplete conversations, including completing the documentation sweep in the API layer.

Importantly, I'd like to see `@returns`, `@params`, etc. in the javadocs.
This does lead to some repetition, but that's the convention across the project.

fourls (changes requested):
Looks really great, functionality is really impressive and I couldn't find any obvious bugs. I have a lot of comments with questions, ideas for renaming/refactoring, and recs for where to put more tests.

The domain of this PR is really complex, confusing, and wide-ranging - we're basically reviewing an entirely new semantic model of Delphi code, and so there's so much that it's difficult to deep-dive on any single thing. In my view, that means that we need to reduce the cognitive noise as much as possible by improving naming/modelling and making sure testing is comprehensive (since we're relying on them to assert most of the functionality is correct).

fourls:
Also, I'm not sure if you've noticed @jgardn3r but there are still a few unresolved conversations from @Cirras 's first round of review.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/LoopExecutingAtMostOnce.html
```diff
@@ -0,0 +1,62 @@
...
+  break; // Noncompliant
+end;
+</pre>
+<pre data-diff-id="2" data-diff-type="noncompliant">
+for var I := 0 to 10 do begin
+  if I = 2 then
+    break // Noncompliant
+  else begin
+    Writeln(I);
+    return; // Noncompliant
```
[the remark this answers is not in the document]
jgardn3r: Oops!! I never really can tell.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantJumpCheckTest.java
```diff
@@ -0,0 +1,372 @@
...
+        Issues.ExpectingNoIssues);
+  }
+
+  @Test
+  void testForLoopAfterConditionalTryFinallyExitShouldNotAddIssue() {
+    doExitTest(
+        List.of(
+            "try",
+            "  A;",
+            "  if B then begin Exit end; // Compliant",
```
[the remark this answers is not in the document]
jgardn3r: I think I was doing that just to be explicit everywhere to say what the focus was. Can remove though if you still think not useful.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
```diff
@@ -0,0 +1,756 @@
...
+  public ControlFlowGraphBuilder visit(IfStatementNode node, ControlFlowGraphBuilder builder) {
+    BuilderBlock after = builder.getCurrentBlock();
+
+    // process `else`
+    BuilderBlock elseBlock = after;
+    if (node.getElseStatement() != null) {
+      StatementNode elseStatement = node.getElseStatement();
+      if (!(elseStatement instanceof IfStatementNode)) {
+        builder.nextBlockTo(after);
+      }
```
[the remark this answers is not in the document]
jgardn3r: The `elseStatement` is visited the next line down. So if it is an `IfStatement` it will go through this visitor again.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
```diff
@@ -0,0 +1,756 @@
...
+   *                        ┌─> true ────┐
+   * `Body` ─> `Condition` ─┴─> false ─┐ └─> after
+   *   ^───────────────────────────────┘
+   * </pre>
+   */
+  @Override
+  public ControlFlowGraphBuilder visit(RepeatStatementNode node, ControlFlowGraphBuilder builder) {
+    BuilderBlock after = builder.getCurrentBlock();
+    // Create a placeholder for the conditional block
+    BuilderBlock loopback = builder.nextBlockTo(after);
```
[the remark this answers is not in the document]
jgardn3r: It is mainly for the ordering of blocks. If we move the body before the condition, we are evaluating things in a different order to the compiler for a start, but it generates a harder to read cfg. It is logically equivalent but the ordering is different:
```text
Starts at B1

B2
0:	NameReferenceNodeImpl         	Foo
T:	RepeatStatementNodeImpl       	repeat
	jumps to: B0(true) B1(false)

B1
0:	NameReferenceNodeImpl         	Bar
	jumps to: B0

B0
	(Exit)
```

Affordances have to be made for both the condition and the body having multiple blocks.

It also ensure that when you read it, it is the natural (reverse) order of how things are evaluated.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/RepeatStatementNodeImpl.java
```diff
@@ -20,13 +20,25 @@
...
 
+  @Override
+  public ExpressionNode getGuardExpression() {
+    return getFirstChildOfType(ExpressionNode.class);
+  }
+
+  @Override
+  public StatementListNode getStatementList() {
+    return getFirstChildOfType(StatementListNode.class);
+  }
```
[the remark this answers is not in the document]
fourls: In #331 `getChild(<index>)` is used, so that's probably the preferred way.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
```diff
@@ -0,0 +1,756 @@
...
+        });
+    return builder;
+  }
+
+  @Override
+  public ControlFlowGraphBuilder visit(WithStatementNode node, ControlFlowGraphBuilder builder) {
+    build(node.getStatement(), builder);
+    builder.nextBlockToCurrent();
+    build(node.getTargets(), builder);
+    builder.nextBlockToCurrent();
```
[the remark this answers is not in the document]
jgardn3r: I think you're right, Ill add it.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantJumpCheckTest.java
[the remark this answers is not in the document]
fourls: I think it's probably best to remove for consistency with the other rules.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    return exit;
+  }
+
+  @Override
+  public List<Block> getBlocks() {
+    return Collections.unmodifiableList(Lists.reverse(blocks));
+  }
+
+  /// Removes redundant blocks from the graph and updates their neighbouring blocks.
+  public ControlFlowGraphImpl pruned() {
```
fourls, line 60: `ControlFlowGraphImpl::pruned` returning a `ControlFlowGraphImpl` makes it seem like the graph is cloned, when it's actually pruned inplace. It might be better to remove the return type and rename to `prune()` for clarity.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
```diff
@@ -0,0 +1,1324 @@
...
+                .withTerminator(IfStatementNode.class),
+            terminator(StatementTerminator.CONTINUE).jumpsTo(1, 1),
+            block(element(NameReferenceNode.class, "I"))
+                .branchesTo(3, 0)
+                .withTerminator(ForInStatementNode.class)));
+  }
+
+  @Test
+  void testBreakOutsideOfLoop() {
+    assertThatThrownBy(() -> test("Break;", checker()))
```
fourls, line 674: SonarLint:

> Refactor the code of the lambda to have only one invocation possibly throwing a runtime exception.

This `checker()` invocation should be moved out of the lambda into a separate variable that is then captured.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
```diff
@@ -0,0 +1,1324 @@
...
+        .withFailMessage("'Break' statement not in loop statement.")
+        .isInstanceOf(IllegalStateException.class);
+  }
+
+  @Test
+  void testContinueOutsideOfLoop() {
+    assertThatThrownBy(() -> test("Continue;", checker()))
```
fourls, line 681: SonarLint:

> Refactor the code of the lambda to have only one invocation possibly throwing a runtime exception.

This `checker()` invocation should be moved out of the lambda into a separate variable that is then captured.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
fourls, line 1: Could we add one control flow graph test that's a number of quite complex constructions? All of these tests are great, but since they test each construct in isolation there's a possibility for bugs slipping between the cracks (e.g. the incorrect grouping/ordering of the `with` blocks).

fourls (changes requested):
Looks good! Fixed most of my concerns from the previous round of review.
This round I had a bit more of a look at the CFG tests to see if everything matched up - mostly good, I have a few comments.

delphi-checks-testkit/src/test/java/au/com/integradev/delphi/checks/verifier/CheckVerifierImplTest.java
```diff
@@ -367,6 +382,217 @@ void testImpliedAndNotImpliedIssue() {
...
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("validSecondaryLocationsCases")
+  void testValidSecondaryLocationsOnIssue(String annotation1, String annotation2) {
+    Supplier<CheckVerifier> verifier =
+        () ->
+            CheckVerifier.newVerifier()
+                .withCheck(
+                    new WillRaiseSecondaryOnAllSubsequentEvenAndPreviousOddIntegerLiteralsCheck())
```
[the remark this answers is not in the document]
jgardn3r: Conceptually and implementation-wise it isn't that complex, but naming the thing some less descriptive felt kind of wrong. Open for ideas on other ways to test/name this.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockBuilder.java
```diff
@@ -0,0 +1,414 @@
...
+public final class ProtoBlockBuilder {
+  private Function<List<DelphiNode>, Block> blockSupplier;
+  private BiConsumer<Map<ProtoBlock, Block>, Block> dataSetter;
+
+  private ProtoBlockBuilder() {
+    this.blockSupplier = TerminusImpl::new;
+    this.dataSetter = (blocks, block) -> {};
+  }
+
+  public static ProtoBlockBuilder newBlock() {
```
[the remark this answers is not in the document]
jgardn3r: The reason it is called newBlock is because, where I use it, it is statically imported.
Which do you prefer?
- `builder.addBlock(new().with...)`
- `builder.addBlock(newBlock().with...)`
- `builder.addBlock(new ProtoBlockBuilder().with...)`
Or removing the static imports
- `builder.addBlock(ProtoBlockBuilder.new().with...)`

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockBuilder.java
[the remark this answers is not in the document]
jgardn3r: I think somewhere in development there was a time where there was chaining. But at this point in time, there isn't a need and you make a good point.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockBuilder.java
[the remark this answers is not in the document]
jgardn3r: Ties into the other comment about the nature of this whole class.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/BlockBuilder.java
```diff
@@ -0,0 +1,414 @@
...
+  public static BlockBuilder newBlock() {
+    return new BlockBuilder();
+  }
+
+  private BlockBuilder() {
+    this.blockSupplier = TerminusImpl::new;
+    this.dataSetter = (blocks, block) -> {};
+  }
+
+  public BlockBuilder withTerminator(DelphiNode terminator) {
```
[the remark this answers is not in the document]
jgardn3r: You do make a good point about the successor. Thinking about it again, I think it might make more sense as its own renamed type. Reading the `Sink` javadoc I wrote again, I also now realise that it is wrong.

`Halt`s have no successors and are a sink in the graph (it is an abrupt process termination, nothing happens after)
`Exit`s "jump" to the end of the routine (and has to be a jump because of finallys and the like)

I think this modelling makes the most sense, so I will go with the ilk of your original comment.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockBuilder.java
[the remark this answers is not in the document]
jgardn3r: I will change the class to a factory and remove all the chaining capabilities.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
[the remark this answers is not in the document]
jgardn3r: I think another factor that I forgot to mention earlier. The block of the repeat can be more than one block.
The loopback (by being updated at the end) acts as a pointer to the first block of the body. 
Additionally, the condition block can also be more than one block.

Does that make more sense? Do you think your thinking still works?

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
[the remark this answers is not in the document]
jgardn3r: After looking around some more, I can't see a good reason for it to be there. I'll guess that's how it is in sonar-java and, as we know, I initially copied it heavily. Without it, it really simplifies the logic.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
```diff
@@ -0,0 +1,1324 @@
...
+  @Test
+  void testIfThenElse() {
+    test(
+        "if A then Foo else Bar;",
+        checker(
+            block(element(NameReferenceNode.class, "A"))
+                .branchesTo(2, 1)
+                .withTerminator(IfStatementNode.class),
+            block(element(NameReferenceNode.class, "Foo")).succeedsTo(0),
+            block(element(NameReferenceNode.class, "Bar")).succeedsTo(0)));
```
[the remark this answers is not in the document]
jgardn3r: The main reason why I am hesitant to do that is, the ID of block is not in the public API. So we are asserting an implementation detail. I do agree that it isn't super clear which order the blocks are in.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  public static ProtoBlock finallyBlock(ProtoBlock successor, ProtoBlock finallySuccessor) {
+    return new ProtoBlock(
+        FinallyImpl::new,
+        (blocks, block) ->
+            ((FinallyImpl) block).setData(blocks.get(successor), blocks.get(finallySuccessor)));
+  }
+
+  public static ProtoBlock succeedingTo(ProtoBlock successor) {
```
fourls, line 66: `succeedingTo` is a little confusing at first glance. I think the most important thing about the block this function creates is that it's totally linear. Maybe `linear`? Or even `before`?

```java
ProtoBlockFactory.succeedingTo(successor)
// vs
ProtoBlockFactory.linear(successor)
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
```diff
@@ -0,0 +1,395 @@
...
+
+  public static ProtoBlock exitBlock() {
+    return new ProtoBlock(TerminusImpl::new, (blocks, block) -> {});
+  }
+
+  public static ProtoBlock halt(DelphiNode terminator) {
+    return new ProtoBlock(HaltImpl::new, (blocks, block) -> ((HaltImpl) block).setData(terminator));
+  }
+
+  public static ProtoBlock withBranch(
```
fourls, line 50: I think this can also have its `with` removed.

```suggestion
  public static ProtoBlock branch(
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
```diff
@@ -0,0 +1,395 @@
...
+      DelphiNode terminator, ProtoBlock trueBlock, ProtoBlock falseBlock) {
+    return new ProtoBlock(
+        BranchImpl::new,
+        (blocks, block) ->
+            ((BranchImpl) block)
+                .setData(terminator, blocks.get(trueBlock), blocks.get(falseBlock)));
+  }
+
+  public static ProtoBlock finallyBlock(ProtoBlock successor, ProtoBlock finallySuccessor) {
```
fourls, line 59: Does this function need to end with `Block`?

```suggestion
  public static ProtoBlock finally(ProtoBlock successor, ProtoBlock finallySuccessor) {
```

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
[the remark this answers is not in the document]
fourls: Ah right, didn't catch that. Is there any particular reason it can't be in the public API? It may be useful to consumers for e.g. uniquely identifying a block.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/checker/BlockChecker.java
```diff
@@ -0,0 +1,321 @@
...
+package au.com.integradev.delphi.cfg.checker;
+
+import static org.assertj.core.api.Assertions.*;
+
+import au.com.integradev.delphi.cfg.api.Block;
+import au.com.integradev.delphi.cfg.api.Branch;
+import au.com.integradev.delphi.cfg.api.Cases;
+import au.com.integradev.delphi.cfg.api.Finally;
+import au.com.integradev.delphi.cfg.api.Linear;
+import au.com.integradev.delphi.cfg.api.Halt;
```
fourls, line 28: These imports are out of order - needs a `mvn spotless:apply`. It looks like the build and format checks aren't running on this PR anymore - they might be out of date since this is a very old branch.

fourls (changes requested):
Looks good! Only a few quite minor things now. Not sure why GitHub's split them up into a few mini reviews.

You'll have to rebase at some point as there are merge conflicts + the workflows have stopped running.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
jgardn3r: `finally` is a keyword so it needs something.

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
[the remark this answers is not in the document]
jgardn3r: https://github.com/integrated-application-development/sonar-delphi/pull/281#discussion_r1802261124
I think erring on the side of it not being there with the possibility of adding it later was discussed here.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Cases.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Set;
+
+/** A block for the {@code case} statement's behaviour of a possible successor for each arm */
+public interface Cases extends Block, Terminated {
+  /**
+   * All the cases this statement can succeed to in the {@link ControlFlowGraph}
+   *
+   * @return the {@link Set<Block>} of {@code case} arm {@link Block}s
+   */
+  Set<Block> getCaseSuccessors();
```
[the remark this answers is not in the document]
Cirras: A nonnull "fallthrough" successor would also make sense.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/ControlFlowGraph.java
```diff
@@ -0,0 +1,45 @@
...
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.List;
+
+/** A graph representation of all paths that could be traversed in the execution of code */
+public interface ControlFlowGraph {
+  /**
+   * The entry block to the control flow graph
+   *
+   * @return the starting {@link Block} of the {@link ControlFlowGraph}
```
Cirras, line 28:
```suggestion
   * @return the starting block of the control flow graph
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/ControlFlowGraph.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  Block getEntryBlock();
+
+  /**
+   * The final exit block of the control flow graph
+   *
+   * @return the final {@link Block} of the {@link ControlFlowGraph}
```
Cirras, line 35:
```suggestion
   * @return the final block of the control flow graph
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/ControlFlowGraph.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  Block getExitBlock();
+
+  /**
+   * All the blocks within the control flow graph
+   *
+   * @return the {@link List<Block>} of all {@link Block}s in the {@link ControlFlowGraph}
```
Cirras, line 42:
```suggestion
   * @return the list of all blocks in the control flow graph
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Finally.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/** A {@code finally} block whose control flow depends on the existence of previous exceptions */
+public interface Finally extends Block {
+  /**
+   * Next block in the {@link ControlFlowGraph} without exceptional circumstances
+   *
+   * @return the successor {@link Block} of the {@code finally} block
```
Cirras, line 30:
```suggestion
   * @return the successor block of the {@code finally} block
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Finally.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  Block getSuccessor();
+
+  /**
+   * Next block in the {@link ControlFlowGraph} if there were exceptional circumstances
+   *
+   * @return the successor {@link Block} if the {@code finally} block was reached by an exception
```
Cirras, line 37:
```suggestion
   * @return the successor block if the {@code finally} block was reached by an exception
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Halt.java
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.Collections;
+import java.util.Set;
+
+/** A block which represents a {@code Halt}, an abrupt process termination */
```
Cirras, line 24:
```suggestion
/** A block which represents a {@code Halt}, which is an immediate process termination */
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Linear.java
```diff
@@ -0,0 +1,36 @@
...
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.Set;
+
+/** A block which has a single successor */
+public interface Linear extends Block {
+  /**
+   * Next block in the {@link ControlFlowGraph}
+   *
+   * @return the successor {@link Block}
```
Cirras, line 28:
```suggestion
   * @return the successor block
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Halt.java
```diff
@@ -0,0 +1,30 @@
...
+public interface Halt extends Block, Terminated {
```
Cirras, line 25: Should this implement `Terminus`, or is there a reason it doesn't?
If it did, we wouldn't need the `getSuccessors` override.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnconditionalJump.java
```diff
@@ -0,0 +1,43 @@
...
+package au.com.integradev.delphi.cfg.api;
+
+import java.util.Set;
+
+/** A block which has particular alteration in control flow, e.g., {@code goto} */
+public interface UnconditionalJump extends Block, Terminated {
+  /**
+   * The jumping target block in the {@link ControlFlowGraph}
+   *
+   * @return the successor {@link Block} of the jump
```
Cirras, line 28:
```suggestion
   * @return the successor block of the jump
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnconditionalJump.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  Block getSuccessor();
+
+  /**
+   * The next block in the {@link ControlFlowGraph} if the jump were to be removed
+   *
+   * @return the successor {@link Block} if this control flow alteration were to be removed
```
Cirras, line 35:
```suggestion
   * @return the successor block if this control flow alteration were to be removed
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnknownException.java
```diff
@@ -0,0 +1,52 @@
...
+ * A block which may alter the control flow with an exception, e.g., routine invocations in a
+ * try-catch
+ *
+ * <p>Known exceptions with be directly specified with a {@code Linear} block to their target
+ */
+public interface UnknownException extends Block {
+  /**
+   * Next block without exceptional circumstances
+   *
+   * @return the successor {@link Block} without exceptions
```
Cirras, line 35:
```suggestion
   * @return the successor block without exceptions
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/UnknownException.java
```diff
@@ -0,0 +1,52 @@
...
+   */
+  Block getSuccessor();
+
+  /**
+   * Possible exit paths in exceptional circumstance
+   *
+   * @return the {@link Set<Block>} of successor {@link Block}s that could be jumped to in the event
+   *     of an exception
```
Cirras, line 43:
```suggestion
   * @return the set of successor blocks that could be jumped to in the event of an exception
```

delphi-frontend/src/test/java/au/com/integradev/delphi/cfg/ControlFlowGraphTest.java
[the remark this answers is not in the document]
fourls: As discussed, the value used in `succeedsTo`/`branchesTo` is the reverse index of the block checker in the list, starting from 1. Although this isn't obvious and could be better exposed in the future, no need to expose the ID for now.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
fourls: Ah, gotcha

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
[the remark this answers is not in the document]
fourls: That's much clearer :+1:

fourls (dismissed):
Looks good to me! This will be a really great new feature that can power a lot of really interesting rules.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/CheckTestNameTest.java
```diff
@@ -137,6 +139,8 @@ void testCheckTestsVerifyingNoIssuesAreNamedCorrectly() {
         .and(not(CALL_ASSERT_THROW_BY))
         .should()
         .haveNameMatching(".*ShouldNotAddIssues?$")
+        .orShould()
+        .notBeAnnotatedWith(Test.class)
```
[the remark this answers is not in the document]
jgardn3r: I just realised that neither option covers parameterised tests. I will add another commit changing the other tests too.

jgardn3r:
Went a little overzealous with the delete button and failed to compile again. All @Cirras comments should be addressed now.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Cases.java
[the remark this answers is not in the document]
jgardn3r: This one isn't addressed, doing that now.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Halt.java
```diff
@@ -0,0 +1,22 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.cfg.api;
+
+/** A block which represents a {@code Halt}, which is an immediate process termination */
```
Cirras, line 21: wording nit
```suggestion
/** A block representing a {@code Halt}, which is an immediate process termination */
```

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/Terminator.java
```diff
@@ -0,0 +1,70 @@
...
+          case "System.Exit":
+            return TerminatorKind.EXIT;
+          case "System.Break":
+            return TerminatorKind.BREAK;
+          case "System.Halt":
+            return TerminatorKind.HALT;
+          case "System.Continue":
+            return TerminatorKind.CONTINUE;
+          default:
+            return TerminatorKind.NODE;
```
Cirras, line 56: Redundant, isn't it? You can let it fall through to the return at the end of the function.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/api/Cases.java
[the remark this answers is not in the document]
Cirras: Don't forget this one.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
Cirras, line 1: Most (all?) of the comments in this file should not be javadoc comments, they seem to be describing implementation details or things that would otherwise not make sense to see in generated javadocs.

Cirras (changes requested):
This time for sure, maybe!

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphVisitor.java
jgardn3r: I guess I did that so it would be rendered. But if you don't want that, I can remove the javadoc-ness of them.

Cirras (dismissed):
I'm happy with this now - well done. 👍

There's a trivial follow-up, which is releasing the API for custom plugins. We'll do that once we've built some more things on this feature and are confident that we like the current shape of the CFG API.

Cirras (approved):

PR #284, opened 2024-08-12
Improve type inference

### LINKED ISSUE #123: Widen real types to `Extended` when inferring the type of an inline var
### Prerequisites

- [X] This improvement would be generally useful, not specific to my code or setup.

### Engine area

Delphi language support

### Improvement description

In the below example, `Bar` is inferred as `Extended` despite the assigned value having an explicitly-known type of `Double`.
```pas
var
  Foo: Double;
begin
  var Bar := Foo;
end;
```
Our type inference rules don't currently reflect this behavior.

### Rationale

Match the behavior of the compiler, even when it's weird.

See: https://quality.embarcadero.com/browse/RSP-42107

Cirras:
This PR generally improves type inference around:
- unsigned integer literals
- array constructor expressions containing integer literals
- real expressions (closes #123)

Additionally, I've added a `TypeInferrerTest` suite to test the `TypeInferrer` more directly, as opposed to the more limited/indirect testing we're currently getting through the `SymbolTableExecutor` tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improve type inference around unsigned integer literals.
 - `out` parameters are treated as uninitialized at the start of a routine in
   `VariableInitialization`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve type inference around unsigned integer literals.
 - Improve type inference around array constructors containing integer literals.
+- Improve type inference around real expressions.
 - `out` parameters are treated as uninitialized at the start of a routine in
   `VariableInitialization`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improve type inference around unsigned integer literals.
+- Improve type inference around array constructors containing integer literals.
 - `out` parameters are treated as uninitialized at the start of a routine in
   `VariableInitialization`.
 
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/TypeInferrer.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/TypeInferrer.java
@@ -50,14 +50,7 @@ public final class TypeInferrer {
       if (arrayConstructor instanceof ArrayConstructorNode) {
         type = inferArrayConstructor((ArrayConstructorNode) arrayConstructor);
       } else {
-        IntegerLiteralNode literal = ExpressionNodeUtils.unwrapInteger(expression);
-        if (literal != null) {
-          IntegerType integer = (IntegerType) typeFactory.getIntrinsic(IntrinsicType.INTEGER);
-          if (integer.min().compareTo(literal.getValue()) <= 0
-              && integer.max().compareTo(literal.getValue()) >= 0) {
-            type = typeFactory.getIntrinsic(IntrinsicType.INTEGER);
-          }
-        }
+        type = widenIntegerLiteralToInteger(type, expression);
       }
     }
 
@@ -67,6 +60,8 @@ public final class TypeInferrer {
         type = returnType;
       }
     }
+
+    type = widenFloatingPointToExtended(type);
 
     return type;
   }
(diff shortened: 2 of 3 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/TypeInferrer.java
```diff
@@ -48,8 +50,15 @@ public Type infer(Typed typed) {
...
         type = inferArrayConstructor((ArrayConstructorNode) arrayConstructor);
-      } else if (ExpressionNodeUtils.isIntegerLiteral(expression) && type.size() <= 4) {
-        type = typeFactory.getIntrinsic(IntrinsicType.INTEGER);
+      } else {
+        IntegerLiteralNode literal = ExpressionNodeUtils.unwrapInteger(expression);
+        if (literal != null) {
+          IntegerType integer = (IntegerType) typeFactory.getIntrinsic(IntrinsicType.INTEGER);
+          if (integer.min().compareTo(literal.getValue()) <= 0
+              && integer.max().compareTo(literal.getValue()) >= 0) {
+            type = typeFactory.getIntrinsic(IntrinsicType.INTEGER);
```
fourls: Should this return an inferred type if it's out of these ranges? e.g. `Cardinal` or `Int64`? I'm not sure how Delphi's inference works here.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improve type inference around unsigned integer literals.
```
fourls, line 16: This is specifically around array constructors, right? This should probably be clarified in the changelog.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improve type inference around unsigned integer literals.
+- Improve type inference around array constructors containing integer literals.
```
fourls, line 17: Could probably be combined with the previous entry.

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/TypeInferrerTest.java
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  private static CommonDelphiNode commonDelphiNode(int type, String text) {
+    return new CommonDelphiNodeImpl(new CommonToken(type, text));
+  }
+
+  private static void assertInferred(ExpressionNode expression, Type type) {
+    doAssertInferred(expression, type);
+    if (expression != null) {
+      doAssertInferred(parenthesize(expression), type);
```
fourls, line 317: What's the reasoning behind this?

fourls (changes requested):
Looks good on the whole! Just some questions.

I will assume the inference logic is correct for Delphi - I haven't tested any of this to see if what you've added matches up with Delphi behaviour.

I did notice a typo in the first commit body: `errneously` should be `erroneously`.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/TypeInferrer.java
Cirras: The weird part of the inference process is just widening anything within the `Integer` range to `Integer`.
Otherwise, we take the type of the expression based on the literal value.
Based on that, the possible integer types are `Integer` `Cardinal`, `Int64`, and `UInt64`. This is reflected in the unit tests.

CHANGELOG.md
Cirras: It's not specifically around array constructors - there were distinct problems with the type inference for standalone integer expressions.

CHANGELOG.md
Cirras: The array constructors type inference issues were a distinct problem from what was fixed with the integer type inference.
The problem with array constructors was the way we determined which element "won" and became the element type for the resulting dynamic array. We were doing a dumb size comparison, without consideration for signedness.

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/TypeInferrerTest.java
Cirras: It ensures we're skipping parentheses in `TypeInferrer`.

Cirras:
Thanks for catching that commit message typo.
Regarding testing, before implementing the changes in this PR I spent some time reproducing these inference rules for integer, real, and array constructor types.

CHANGELOG.md
fourls: Ah, I missed the `TypeInferrer::infer` usage in `VariableNameDeclarationImpl`. Makes sense.

fourls (approved):
OK, looks good to me! I expect this should fix a few sneaky name resolution problems.

Issue #62, opened 2023-10-31
New rule: Imports should be fully qualified


fourls:
### Prerequisites

- [X] This rule has not already been suggested.
- [X] This should be a new rule, not an improvement to an existing rule.
- [X] This rule would be generally useful, not specific to my code or setup.

### Suggested rule title

Imports should be fully qualified

### Rule description

This rule would detect unit imports that are a subset of the actual unit name. For example,

```delphi
uses SysUtils;
```

would raise an issue, but

```delphi
uses System.SysUtils;
```

wouldn't.

### Rationale

There are a number of benefits to fully qualifying unit names, including

* No ambiguity about which unit is being imported
* Consistency across the codebase
* Improved compile time due to not having to look up implicit scopes

A project could remove unit scopes from their `.dproj`, causing the compiler to be unable to resolve unqualified unit names. This is not always feasible, most obviously when the project compiles third-party source code as part of the compilation process - in times like these a Sonar rule would be useful.

Cirras:
Good suggestion, and this would be pretty easy to implement:
* visit `UnitImportNode`
* look at the import declaration (`UnitImportNode::getImportNameDeclaration`)
* compare the length against the original unit declaration (`UnitImportNameDeclaration::getOriginalDeclaration`)

PR #287, opened 2024-08-21
Create rule FullyQualifiedImportCheck

EduardoVaz06:
This PR implements a new rule: FullyQualifiedImportCheck.

As sugested in https://github.com/integrated-application-development/sonar-delphi/issues/62, this rule simply checks if the imports in the UsesClausesNode are fully qualified. If they are not, an issue is raised, suggesting to change it to be fully qualified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `FullyQualifiedImportCheck` analysis rule, which flags non-fully qualified imports.
 - **API:** `CaseItemStatementNode::getExpressions` method.
 
 ### Changed
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
@@ -1,0 +1,59 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2024 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.communitydelphi.api.ast.UnitImportNode;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.reporting.QuickFix;
+import org.sonar.plugins.communitydelphi.api.reporting.QuickFixEdit;
+import org.sonar.plugins.communitydelphi.api.symbol.declaration.UnitImportNameDeclaration;
+
+@Rule(key = "FullyQualifiedImport")
+public class FullyQualifiedImportCheck extends DelphiCheck {
+
+  @Override
+  public DelphiCheckContext visit(UnitImportNode unitImportNode, DelphiCheckContext context) {
+    if (!unitImportNode.isResolvedImport()) {
+      return context;
+    }
+
+    UnitImportNameDeclaration unitImportNameDeclaration = unitImportNode.getImportNameDeclaration();
+
+    String unitFullyQualifiedName = unitImportNameDeclaration.getOriginalDeclaration().getImage();
+    String unitImportName = unitImportNameDeclaration.getImage();
+
+    if (unitImportName.length() != unitFullyQualifiedName.length()) {
+      context
+          .newIssue()
+          .onNode(unitImportNode)
+          .withMessage(
+              "Fully qualify this unit name (found: \"%s\" expected: \"%s\").",
+              unitImportName, unitFullyQualifiedName)
+          .withQuickFixes(
+              QuickFix.newFix("Fully qualify unit import")
+                  .withEdit(
+                      QuickFixEdit.replace(unitImportNode.getNameNode(), unitFullyQualifiedName)))
+          .report();
+    }
+
+    return context;
+  }
+}
--- a/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
+++ b/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
@@ -1,0 +1,21 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Imports should be done using the full name of the unit, as this facilitates code readability, avoids potential ambiguities about which unit is being imported, and ensures consistency throughout the codebase.
+</p>
+<p>
+  Using the full name of the units not only improves code clarity and maintenance but also optimizes the compilation process. The compiler does not need to spend time searching implicit scopes to resolve which units are being referenced. By specifying the full unit name, the compiler can immediately locate the necessary files, resulting in faster and more efficient compilation.
+</p>
+<h2>How to fix it</h2>
+<p>
+  Fully qualify the unit name.
+</p>
+<pre data-diff-id="1" data-diff-type="noncompliant">
+uses
+  Classes,
+  SysUtils;
+</pre>
+<pre data-diff-id="1" data-diff-type="compliant">
+uses
+  System.Classes,
+  System.SysUtils;
+</pre>
--- a/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
+++ b/delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
@@ -1,0 +1,18 @@
+{
+  "title": "Imports should be fully qualified",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant/Issue",
+    "constantCost": "5min"
+  },
+  "code": {
+    "attribute": "CLEAR",
+    "impacts": {
+      "MAINTAINABILITY": "LOW"
+    }
+  },
+  "tags": ["convention"],
+  "defaultSeverity": "Minor",
+  "scope": "ALL"
+}
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/CheckList.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/CheckList.java
@@ -83,6 +83,7 @@ public final class CheckList {
           FormatArgumentTypeCheck.class,
           FormatStringValidCheck.class,
           FreeAndNilTObjectCheck.class,
+          FullyQualifiedImportCheck.class,
           GotoStatementCheck.class,
           GroupedFieldDeclarationCheck.class,
           GroupedParameterDeclarationCheck.class,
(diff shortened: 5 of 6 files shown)


delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
```diff
@@ -0,0 +1,18 @@
...
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "code": {
+    "attribute": "CLEAR",
+    "impacts": {
+      "MAINTAINABILITY": "LOW"
+    }
+  },
+  "tags": [],
```
fourls, line 15: Every rule should have at least one tag. We tend to use the same tags used in SonarSource's official plugins, which can be found [here](https://docs.sonarsource.com/sonarqube/latest/user-guide/rules/built-in-rule-tags/?gad_source=1). 

In this case, probably just `convention` is fine.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
```diff
@@ -0,0 +1,32 @@
...
+    Imports should be done using the full name of the unit, as this facilitates code readability, avoids potential ambiguities about which unit is being imported, and ensures consistency throughout the codebase.
+</p>
+<p>
+    Using the full name of the units not only improves code clarity and maintenance but also optimizes the compilation process. The compiler does not need to spend time searching implicit scopes to resolve which units are being referenced. By specifying the full unit name, the compiler can immediately locate the necessary files, resulting in faster and more efficient compilation.
+</p>
+<h2>How to fix it</h2>
+<p>
+    Fix the import to use the full name of the unit.
+</p>
+<h3>Non-compliant Code Example</h3>
```
fourls, line 12: We don't tend to use the "Non-compliant code example" headers, since the diff view of the code is usually explanatory enough. Would you be able to remove them?

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
```diff
@@ -0,0 +1,32 @@
...
+<pre data-diff-id="1" data-diff-type="noncompliant">
+unit VariableUtilsU;
+
+interface
+
+uses
+  Classes,
+  SysUtils;
```
fourls, line 20: There's no need to include the top of the unit, especially with this arbitrary unit name - we can probably just show the uses clause. 

```suggestion
uses
  Classes,
  SysUtils;
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
```diff
@@ -0,0 +1,58 @@
...
+
+    String unitFullyQualifiedName = unitImportNameDeclaration.getOriginalDeclaration().getImage();
+    String unitImportName = unitImportNameDeclaration.getImage();
+
+    if (unitImportName.length() != unitFullyQualifiedName.length()) {
+      context
+          .newIssue()
+          .onNode(unitImportNode)
+          .withMessage(
+              "Specify the fully qualified unit name (found: \"%s\" expected: \"%s\").",
```
fourls, line 48: Issue messages should use *this* instead of *the*:

```suggestion
              "Fully qualify this unit name (found: \"%s\" expected: \"%s\").",
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
```diff
@@ -0,0 +1,58 @@
...
+              unitImportName, unitFullyQualifiedName)
+          .withQuickFixes(
+              QuickFix.newFix("Fix to \"%s\"", unitFullyQualifiedName)
```
fourls, line 51: Tiny nitpick, it's not technically _fixing_ the import since it was still OK in its previous iteration. 
How about something like this?

```suggestion
              QuickFix.newFix("Fully qualify unit import")
```

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
```diff
@@ -0,0 +1,18 @@
+{
+  "title": "Imports should be fully qualified",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
```
fourls, line 6: No need to escape this forward slash. 
```suggestion
    "func": "Constant/Issue",
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
```diff
@@ -0,0 +1,58 @@
...
+                  .withEdit(QuickFixEdit.replace(unitImportNode, unitFullyQualifiedName)))
```
fourls, line 52: This would have the wrong effect when quick fixing unit imports with the `in` keyword:

```delphi
uses
  Unit in "My.Unit.pas";
```

would overwrite the `in "MyUnit.pas"` and become 

```delphi
uses
  My.Unit;
```

This should work:

```suggestion
                  .withEdit(QuickFixEdit.replace(unitImportNode.getNameNode(), unitFullyQualifiedName)))
```

Could you add a test for the `x in "y.pas"` case?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `FullyQualifiedImportCheck` rule, wich analyzes if imports are fully qualified
```
fourls, line 12: To make it more regular with our other changelog entries, how about:

```suggestion
- `FullyQualifiedImportCheck` analysis rule, which flags non-fully qualified imports.
```

fourls (changes requested):
Hi @EduardoVaz06, thanks for this contribution! Looks great and will be a useful rule for codebases looking to gradually transition to fully qualified imports.

I've made a few comments below, nothing major. Also note that the failing CI is just complaining about the rule having no tags.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
EduardoVaz06: Done!

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
EduardoVaz06: Of course! Done

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
EduardoVaz06: Done!

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
EduardoVaz06: Done!

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
EduardoVaz06: Makes total sense! Done!

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.json
EduardoVaz06: Done!

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
EduardoVaz06: Ok!

I added the test, but I'm having trouble making it work. I keep getting 'Unable to build Delphi File" due to the "Unit in 'UnitU.pas'"...

CHANGELOG.md
EduardoVaz06: Done!

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
zaneduffield: If you copied the suggestion from @fourls directly, it's probably because you've got double quotes around the filename when it's expecting single quotes.

zaneduffield: I think the `in` syntax for the `uses` clause can only be used for a `program` file, not a `unit`.

I'm not sure if there is a `DelphiTestProgramBuilder` but if there is then that's what you want to use.

zaneduffield: https://github.com/integrated-application-development/sonar-delphi/blob/bb975dec84521054642a1431103de3ad4478d365/delphi-checks-testkit/src/main/java/au/com/integradev/delphi/builders/DelphiTestProgramBuilder.java

fourls: Oops, ignore my accidental double quotes... 

Yes, as you say `in` does not work for normal units - `DelphiTestProgramBuilder` should work.

Cirras: @EduardoVaz06 did you see these responses a couple weeks ago?
Your test error should be pretty easy to fix with `DelphiTestProgramBuilder`.

EduardoVaz06: Hello! Sorry for the huge delay, I had some personal issues, but I just added the tests!

EduardoVaz06: I noticed that format check has failed, is there something else I need to do?

Cirras: Hi @EduardoVaz06, the project enforces `google-java-format` styling.
If you run `mvn install`, your code will automatically be reformatted correctly.

EduardoVaz06: Done, @Cirras!

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
```diff
@@ -0,0 +1,21 @@
...
+<pre data-diff-id="1" data-diff-type="noncompliant">
+uses
+  Classes,
+  SysUtils;
+</pre>
+<pre data-diff-id="1" data-diff-type="compliant">
+uses
+  System.Classes,
+  System.SysUtils;
+</pre>
```
Cirras, line 21: Can we use 2-space indentation here?
```suggestion
<h2>Why is this an issue?</h2>
<p>
  Imports should be done using the full name of the unit, as this facilitates code readability, avoids potential ambiguities about which unit is being imported, and ensures consistency throughout the codebase.
</p>
<p>
  Using the full name of the units not only improves code clarity and maintenance but also optimizes the compilation process. The compiler does not need to spend time searching implicit scopes to resolve which units are being referenced. By specifying the full unit name, the compiler can immediately locate the necessary files, resulting in faster and more efficient compilation.
</p>
<h2>How to fix it</h2>
<p>
  Fully qualify the unit name.
</p>
<pre data-diff-id="1" data-diff-type="noncompliant">
uses
  Classes,
  SysUtils;
</pre>
<pre data-diff-id="1" data-diff-type="compliant">
uses
  System.Classes,
  System.SysUtils;
</pre>
```

Cirras (changes requested):
One last thing to tweak, just formatting consistency in the rule's HTML description.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/FullyQualifiedImport.html
EduardoVaz06: Done!

Cirras (approved):
Thanks for your contribution!

PR #291, opened 2024-09-09
Exclude non-trivial `inherited` expressions in `RedundantParentheses`

Cirras:
This PR changes the `RedundantParentheses` rule to exclude non-trivial `inherited` expressions.
Parenthesized bare `inherited` is still flagged, but anything more complex is now ignored.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantParenthesesCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantParenthesesCheck.java
@@ -19,8 +19,10 @@
 package au.com.integradev.delphi.checks;
 
 import org.sonar.check.Rule;
+import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.ParenthesizedExpressionNode;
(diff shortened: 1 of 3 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantParenthesesCheck.java
```diff
@@ -50,4 +51,15 @@ public DelphiCheckContext visit(
     }
     return super.visit(expression, context);
   }
+
+  private static boolean isRedundant(ParenthesizedExpressionNode expression) {
+    ExpressionNode parenthesized = expression.getExpression();
+
+    if (parenthesized instanceof PrimaryExpressionNode) {
+      return !(ExpressionNodeUtils.isInherited(parenthesized)
+          && !ExpressionNodeUtils.isBareInherited(parenthesized));
```
jgardn3r, line 60: Do you think this might be more clear de Morgan'd?

Cirras: Pulled it out into a local static method, what do you think?

jgardn3r: I think that makes sense.

Issue #261, opened 2024-06-18
Unify lexing/preprocessing/parsing into hand-rolled scanner


Cirras:
### Prerequisites

- [X] This improvement has not already been suggested.
- [X] This improvement would be generally useful, not specific to my code or setup.

### Engine area

Delphi language support

## Background

- The current parsing is implemented in 3 steps:
  - Lexing (`DelphiLexer` generated by ANTLR)
  - Preprocessing (`DelphiPreprocessor`)
  - Parsing (`DelphiParser`  generated by ANTLR)
- In addition, we have a hand-rolled `CompilerDirectiveParser` with corresponding `ExpressionLexer` and `ExpressionParser` for handling constant expressions within `{$if}` and `{$elseif}` directives.

## Problem

While this approach keeps the concerns cleanly separated and sounds nice on paper, it doesn't reflect reality at all.

Delphi's parsing is a fairly streamlined single-pass implementation that parses high-level AST structures while reading tokens on-demand from the input data and processing compiler directives as they appear. In other words, it's all mashed together into a single process.

There are many subtle deficiencies in the current parsing that we can trace back to this fundamental difference in how the process is modeled.

## Solution

Merge the 3 parsing steps into a hand-rolled `DelphiScanner`:
  - Lex tokens on-demand instead of doing it upfront.
  - Evaluate compiler directives as they're lexed from the input data instead of afterward.
  - Perform symbol table construction simultaneously instead of afterward.

### Rationale

Enable full support for:
- caret notation (#111)
- compiler directives nested in directive expressions (#260)
- symbol information in directive expressions
  - references to constants
  - `Declared` function
- "vexing" parses that require symbol information to resolve
  ```pas
  // [[OPTION 1]]
  // In this scenario, `Write` is called with 2 arguments
  // parsed as binary expressions: `Foo < T` and `T > (B)` 
  {
    type
      A = class
        procedure Bar;
      end;
     
    const
      Foo = 0;
      T = 0;
      B = 0;
  }

  // [[OPTION 2]]
  // In this scenario, `Write` is called with 1 argument
  // parsed as an invocation of generic function `Foo`.
  {  
    type
      A = class
        function Foo<U, V>(X: Integer): Integer;
        procedure Bar;
      end;
     
      T = class end;
     
    const
      B = 0;
  }

  procedure A.Bar;
  begin
    Write(Foo<T, T>(B));
  end;
  ```

Make it easier to implement:
- constant expression evaluation (#116)

PR #292, opened 2024-09-09
Disambiguate const array literals from parenthesized primary expressions

Cirras:
This PR fixes a parsing bug where const array literal expressions would be confused with parenthesized primary expressions.

The official parser builds type information incrementally during parsing and relies on it to know what may appear on the right side of an assignment (which is another point in favor of #261).

With our current approach to parsing, the fix for this problem is a slightly gnarly repetition of rules in the ANTLR3 grammar.

### DIFF
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantParenthesesCheckTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantParenthesesCheckTest.java
@@ -89,7 +89,7 @@ class RedundantParenthesesCheckTest {
   }
 
   @Test
-  void testParenthesesOnNonTrivialInheritedExpressionShouldAddIssue() {
+  void testParenthesesOnNonTrivialInheritedExpressionShouldNotAddIssue() {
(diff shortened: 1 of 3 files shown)


delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantParenthesesCheckTest.java
jgardn3r, line 1: I think it might be worth adding tests for nested parens inside the const array.

jgardn3r (changes requested):

delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantParenthesesCheckTest.java
Cirras: Good idea - I've added the splendidly-named `testParenthesesOnPrimaryExpressionWithinSingleElementArrayLiteralShouldAddIssue`.

jgardn3r (commented):
Looks good post rebase.

jgardn3r (approved):
Looks good!

PR #294, opened 2024-09-26
Handle aliased imports in `FullyQualifiedImport`

Cirras:
This PR fixes an issue in `FullyQualifiedImport` where unit aliases weren't considered, causing false positives on aliased imports.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
@@ -25,32 +25,32 @@ import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
 import org.sonar.plugins.communitydelphi.api.reporting.QuickFix;
 import org.sonar.plugins.communitydelphi.api.reporting.QuickFixEdit;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.UnitImportNameDeclaration;
+import org.sonar.plugins.communitydelphi.api.symbol.declaration.UnitNameDeclaration;
 
 @Rule(key = "FullyQualifiedImport")
 public class FullyQualifiedImportCheck extends DelphiCheck {
 
   @Override
   public DelphiCheckContext visit(UnitImportNode unitImportNode, DelphiCheckContext context) {
-    if (!unitImportNode.isResolvedImport()) {
+    UnitImportNameDeclaration importDeclaration = unitImportNode.getImportNameDeclaration();
+    UnitNameDeclaration unitDeclaration = importDeclaration.getOriginalDeclaration();
+
+    if (unitDeclaration == null || importDeclaration.isAlias()) {
       return context;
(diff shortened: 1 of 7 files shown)


delphi-checks/src/test/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheckTest.java
```diff
@@ -82,6 +82,18 @@ void testUnresolvedImportShouldNotAddIssue() {
...
   }
 
+  @Test
+  void testUnitAliasImportShouldNotAddIssue() {
+    var testFile = new DelphiTestUnitBuilder().appendDecl("uses").appendDecl("  AliasName;");
+
+    CheckVerifier.newVerifier()
+        .withCheck(new FullyQualifiedImportCheck())
+        .withSearchPathUnit(new DelphiTestUnitBuilder().unitName("Scope.RealName"))
+        .withUnitAlias("AliasName", "Scope.RealName")
```
zaneduffield, line 92: I think this test case could be a bit stronger. The problem I see is that the test will continue to pass if the unit alias isn't working at all; the import will be unresolved and no issue will be raised.

You could *also* add the unit scope name of `Scope` alongside a new unit `Scope.AliasName`, which will cause an issue to be raised if the alias is not working (but the Check *is*).

```suggestion
        .withSearchPathUnit(new DelphiTestUnitBuilder().unitName("Scope.RealName"))
        .withSearchPathUnit(new DelphiTestUnitBuilder().unitName("Scope.AliasName"))
        .withUnitScopeName("Scope")
        .withUnitAlias("AliasName", "Scope.RealName")
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
```diff
@@ -25,35 +25,40 @@
...
+                  .withEdit(QuickFixEdit.replace(unitImportNode.getNameNode(), expected)))
           .report();
     }
 
     return context;
   }
+
+  private static boolean isAliasImport(
+      UnitImportNameDeclaration importDeclaration, UnitNameDeclaration unitDeclaration) {
+    return !importDeclaration.simpleName().equalsIgnoreCase(unitDeclaration.simpleName());
```
zaneduffield, line 62: What if you alias `Scope.Name` to `Name`?

zaneduffield (changes requested):
One extra edge case to consider, and a possible improvement to the test.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/FullyQualifiedImportCheck.java
Cirras: Uh oh, this is a bit of an annoying headache to detect.
Guess it's time for a new API method, like `UnitImportNameDeclaration::isAliased`.

zaneduffield (approved):
Looks good! 👍

Issue #297, opened 2024-10-01
Error raised when analyzing varargs method call


FlavienQui:
### Prerequisites

- [X] This bug is in SonarDelphi, not SonarQube or my Delphi code.
- [X] This bug has not already been reported.

### SonarDelphi version

1.6.0

### Issue description

Running analysis with DelphiLint 1.1.1 in Standalone mode, an error is raised when the code contains a method call with the varargs directive.
![Capture](https://github.com/user-attachments/assets/495ff1d0-d69d-4160-9d64-206c5d8cb109)

### Steps to reproduce

Analyze the file with the minimal Delphi code joined to this issue.

### Minimal Delphi code exhibiting the issue

```
unit Unit1;

interface

procedure MyProcedure;

var
  SetValues: procedure(AValues: Integer); cdecl varargs;

implementation

procedure MyProcedure;
var
  LValue: Integer;
begin
  LValue := 1;
  SetValues(LValue, LValue);
end;

end.
```

Cirras:
Hi @FlavienQui,

The exception is getting raised from the `VariableInitialization` rule, but the actual problem goes a bit deeper into the analyzer's semantic model. It looks like we don't consistently persist `varargs` when generating type information for procedurals.

Thanks for reporting this bug, your minimal example was very helpful.

PR #298, opened 2024-10-02
Improve procedural type modeling around routine directives

### LINKED ISSUE #172: File-level scan failure on anonymous method declaration with a calling convention
### Prerequisites

### SonarDelphi version

1.1.0

### SonarQube version

_No response_

### Issue description

Delphi secretly allows you to specify the calling convention for an anonymous method:

```delphi
Foo := function(const P1: Integer): Boolean stdcall
begin
  Result := True;
end;
```

SonarDelphi's grammar does not support this. There are occurrences of this pattern in the FMX source code, resulting in parsing errors when scanning code that references the `FMX.WebBrowser.Win` unit.

FWIW I'm not sure whether this has any effect or not - I haven't ever seen any reference to this, either officially or in the community.

### Steps to reproduce

Scan the below Delphi code, or a unit including `FMX.WebBrowser.Win`. 

```delphi
unit MyUnit;

interface

implementation

uses System.SysUtils;

type
  TStdFunc<T1, T2> = reference to function(const A: T1): T2 stdcall;

var
  Foo: TStdFunc<Integer, Boolean>;

begin
  // Cannot be assigned to TFunc<Integer, Boolean> because calling convention is different
  Foo := function(const P1: Integer): Boolean stdcall
  begin
    Result := True;
  end;

  WriteLn(Foo(1));
end.
```

Cirras:
This PR makes a few improvements:
* fixes parsing errors on anonymous methods with routine directives
* fixes an issue where non-routine procedural types (procvars, anonymous methods, etc.) could never be considered variadic
  * which fixes a downstream `IndexOutOfBoundsException` that surfaced in the `VariableInitialization` rule 
* tidies the AST a bit around anonymous methods by introducing `AnonymousMethodHeadingNode`
* persists routine directives into `ProcedureType`

Fixes #172 and #297.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **API:** `AnonymousMethodNode::getAnonymousMethodHeading` method.
+- **API:** `AnonymousMethodNode::getDirectives` method.
+- **API:** `AnonymousMethodNode::hasDirective` method.
+- **API:** `AnonymousMethodHeadingNode` node type.
+- **API:** `DelphiTokenType.ANONYMOUS_METHOD` token type.
+- **API:** `DelphiTokenType.ANONYMOUS_METHOD_HEADING` token type.
+
 ### Fixed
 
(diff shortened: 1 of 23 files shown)


fourls (approved):
Looks good! It's quite a simple change even though it's over a few files.

PR #300, opened 2024-10-08
Fix parsing errors on `implementation` keywords in conditional branches

### LINKED ISSUE #299: Parsing error when interface section has invalid conditional blocks
### Prerequisites

- [x] This bug has not already been reported.

### SonarDelphi version

1.8.0

### SonarQube version

_No response_

### Issue description

(originally raised by @Scharrels in integrated-application-development/delphilint#65)

SonarDelphi is unable to parse units with all three of the following characteristics:
* There is invalid code in a conditional block in the interface section,
* The `implementation` keyword is inside a conditional branch, and
* The unit is being processed at an interface-level (i.e. it is not one of the files being fully analysed)

Initial analysis by @zaneduffield suggests it is related to SonarDelphi's partial indexing for interface sections.

### Steps to reproduce

The below unit needs to be included into a file that is being analysed, without itself being part of the analysis. 

Minimal example by @zaneduffield: 

```delphi
unit Bad;

interface

{$IF False}
// inside the not-taken branch
type
  Foo = {$IF False} asdf {$ENDIF} class end;
                    // ^ anything invalid
// 'implementation' must be inside the conditional branch
implementation
{$ELSE}
implementation
{$ENDIF}

end.
```

[debug.zip](https://github.com/user-attachments/files/17285229/debug.zip)

Cirras:
In 50726816e318afb456ad0d04fcb2acb22cdf1478, we implemented keyword case-insensitivity directly into the grammar. This gave us a small performance boost, but quietly impaired our ability to track `directiveNesting` while lexing a source file.

We were using the case-sensitive `String#startsWith` to check for a particular compiler directive while lexing the token, which used to work because the input actually was lowercase before. Now, we need to do a case-insensitive comparison because the input is using the original casing.

This `directiveNesting` issue introduced some subtle issues in cases where:
- we're only parsing the interface of a file (skipping the `implementation` section
- `implementation` keywords are nested within conditional directives (`$IF(DEF)`, `$SLSE`)
- the conditional directives are not lowercase

Fixes #299.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Parsing errors on `implementation` keywords nested in conditional branches.
+
 ## [1.10.0] - 2024-10-01
 
 ### Added
--- a/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
+++ b/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
@@ -122,6 +122,7 @@ import au.com.integradev.delphi.antlr.ast.node.*;
 
 package au.com.integradev.delphi.antlr;
 
+import org.apache.commons.lang3.StringUtils;
 }
 
 @lexer::members {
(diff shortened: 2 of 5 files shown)


fourls (approved):
I'm always suspicious when I can't find something to nitpick about... but looks good to me!

fourls (changes requested):
Aha, found one! I think we need a changelog entry.

fourls (approved):
Looks good.

PR #308, opened 2024-10-28
Reduce duplication around `getDirectives` methods

Cirras:
There's a bit of shared (copy-pasted) code between `AnonymousMethodHeadingNodeImpl` and `RoutineHeadingNodeImpl` around constructing a set of routine directives based on child nodes.

This PR pulls that shared code out into a small util.


fourls (approved):
To me, looks good.

PR #310, opened 2024-11-01
Fix MissingRaise false positives on Self constructor calls

fourls:
This PR fixes an issue with the new MissingRaise rule that would cause it to raise false positives on constructor calls on `Self`, when `Self` is an exception type. For example:

```delphi
type
  EMyError = class(Exception)
  public
    constructor Create(S: string);
    constructor Create(A: Integer; B: Integer);
  end;

constructor EMyError.Create(A: Integer; B: Integer);
begin
  // ...
  Create('Default message'); // <- Raises a false positive issue
  // ...
end;
```

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/MissingRaiseCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/MissingRaiseCheck.java
@@ -55,30 +55,54 @@ public class MissingRaiseCheck extends DelphiCheck {
     return context;
   }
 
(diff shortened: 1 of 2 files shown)


zaneduffield (approved):
Looks good!

It's annoying that Delphi lets you call a constructor method to re-initialise an existing object, but it's not so bad when you realise that all new-object constructor calls are qualified.

zaneduffield:
It's also worth noting that this doesn't need a changelog entry because the bug was never released.

PR #311, opened 2024-11-06
Add an ignore list to MixedNames

EduardoVaz06:
This PR adds a setting on MixedNames rule to ignore certain names.

This PR implements a suggestion I have for MixedNames rule. In my company, we have certain patterns in names, so in some cases it's suggested by the rule to comply with the definition, and ends up not being compliant with our pattern. I think it would be good to have a list for names to be ignored.

For example, sometimes the rules suggest to be WParam, and sometimes WPARAM, due to the definition. If we have a pattern to always be WParam, this addition could help to ignore WParam name, and don't raise any issues.

It's up to you to know if this is a desired feature or not.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/MixedNamesCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/MixedNamesCheck.java
@@ -22,9 +22,13 @@
  */
 package au.com.integradev.delphi.checks;
 
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableSortedSet;
 import java.util.List;
+import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
 import org.sonar.plugins.communitydelphi.api.ast.ArgumentListNode;
 import org.sonar.plugins.communitydelphi.api.ast.AttributeNode;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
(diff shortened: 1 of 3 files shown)


delphi-checks/src/test/java/au/com/integradev/delphi/checks/MixedNamesCheckTest.java
```diff
@@ -23,8 +23,30 @@
...
+        .withCheck(createCheck())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendImpl("procedure Test;")
+                .appendImpl("var")
+                .appendImpl("  b_IGNOREBOOL: Boolean;")
+                .appendImpl("begin")
+                .appendImpl("  b_IgnoreBool := True;")
+                .appendImpl("end;"))
+        .verifyNoIssues();
```
fourls, line 47: Could you add a test that asserts that the exclusion check is not case-sensitive? e.g. having the excluded name be `MyIdentifierName`, but the declaration be `MYIDENTIFIERNAME` and the usage be `myidentifiername`.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/MixedNamesCheckTest.java
```diff
@@ -23,8 +23,30 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
 
 class MixedNamesCheckTest {
+  private static DelphiCheck createCheck() {
+    MixedNamesCheck check = new MixedNamesCheck();
+    check.excludedNames = "b_IgnoreBool";
```
fourls, line 31: I'm not sure what this `b_` prefix is here - is it a Hungarian notation for boolean variables?

Since identifiers with this prefix is unconventional and would be caught by `ClassName`, `VariableName`, etc. it might make people reading this code wonder if there's a specific reason we're testing with this prefix.  Could we change this variable name to something else?

fourls (changes requested):
Hi @EduardoVaz06, thanks for making this contribution! I agree with your reasoning and think it's definitely appropriate to include in the core rule behaviour.

Implementation looks great, no problems I can see. I have a couple of suggestions for some extra tests.

Cirras:
Hi @EduardoVaz06, this PR is a month old.
Do you need anything from us to address the current review feedback?

EduardoVaz06:
> Hi @EduardoVaz06, this PR is a month old. Do you need anything from us to address the current review feedback?

Hi! Sorry for the huge delay. Soon i will make the changes requested and go on with de pr!

delphi-checks/src/test/java/au/com/integradev/delphi/checks/MixedNamesCheckTest.java
Cirras: Done. See `testDifferentlyCasedNameFromExcludedListShouldNotAddIssue`.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/MixedNamesCheckTest.java
Cirras: Done.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/MixedNamesCheck.java
```diff
@@ -48,12 +52,30 @@ public class MixedNamesCheck extends DelphiCheck {
   private static final String MESSAGE = "Avoid mixing names (found: \"%s\" expected: \"%s\").";
   private static final String QUICK_FIX_MESSAGE = "Correct to \"%s\"";
 
+  private Set<String> excludedSet;
+
+  @RuleProperty(
+      key = "excludedNames",
+      description = "List of names to ignore, separated by a comma.")
+  public String excludedNames = "";
```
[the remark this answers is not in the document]
Cirras: Done. See `testNamesInExcludedListShouldNotAddIssue`.

fourls (approved):
Looks good!

PR #312, opened 2024-11-12
Fix return types for `Length`/ `High`/ `Low` on open/dynamic arrays

### LINKED ISSUE #296: Incorrect intrinsic return type for `System.Length` on dynamic arrays
### Prerequisites

### SonarDelphi version

1.9.0

### SonarQube version

_No response_

### Issue description

In Delphi 12 the length of open array parameters was changed to be a `NativeInt` (from `Integer`).
#142 made a change to reflect this, however we didn't realise that this was a fix to bring open arrays in line with dynamic arrays.

Furthermore, the actual return type of the `Length` intrinsic is platform dependent, and not *just* in the way that `NativeInt` is.

Here's a Delphi program that can be used to inspect the return type of `Length`. For good measure, I included const arrays.
```delphi
program LengthIntrinsicType;

{$APPTYPE console}

uses System.TypInfo;

type
  TypeInformation = record
    class function name<T> (const value : T) : string; static;
  end;

class function TypeInformation.name<T> (const value : T) : string;
begin
  Result := GetTypeName(TypeInfo(T));
end;

procedure Test(arr: array of byte);
const
  constarr: array[0..0] of byte = (0);
var
  dynarr: array of byte;
begin
  WriteLn('Dynamic array length type: ', TypeInformation.Name(Length(dynarr)));
  WriteLn('Open array length type: ', TypeInformation.name(Length(arr)));
  WriteLn('Const array length type: ', TypeInformation.name(Length(constarr)));

  writeln(nativeint(typeinfo(nativeint)));
  writeln(nativeint(typeinfo(integer)));
end;

begin
  Test([0]);
  readln;
end.
```

My findings are
| Delphi Version | Platform | Type(Length(OpenArray)) | Type(Length(DynArray)) | Type(Length(ConstArray))
|--------|--------|--------|--------|--------|
| 11 | 32-bit | Integer | Integer | Integer |
| 11 | 64-bit | Integer | NativeInt | Integer |
| 12 | 32-bit | NativeInt | NativeInt | Integer |
| 12 | 64-bit | NativeInt | NativeInt | Integer |

Note that on Delphi 12 NativeInt is a weak alias, so the runtime type information cannot distinguish the alias from the aliased type - this is why my test program prints the `TypeInfo` pointer. To make the results easier to understand, I have substituted `NativeInt` for some of the Delphi 12 results. 

My conclusions are
1. In Delphi 12+ open and dynamic arrays both always have a length of type `NativeInt` (which is a weak alias)
2. In Delphi 11-
    1. Only dynamic arrays have a length of type `NativeInt` on 64-bit but `Integer` on 32-bit (*not* `NativeInt` strongly aliased to `Integer`)
    2. Open arrays always have a length of type `Integer`
3. Const arrays always have a length of `Integer`

The current modelling doesn't match this
https://github.com/integrated-application-development/sonar-delphi/blob/51412489dc8158268ae5e9d7681e92acbe2a6269/delphi-frontend/src/main/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnType.java#L115-L125

### Steps to reproduce

n/a

_No response_

Cirras:
This PR makes the return types for the `Length`, `High`, and `Low` intrinsics more accurate.

The new rules are:
* all array types are always size `Integer` on 32-bit on all compiler versions
* dynamic array types are size `NativeInt` on 64-bit on all compiler versions
* open array types are size `Integer` on 64-bit on all compiler versions prior to 36.0
* open array types are size `NativeInt` on 64-bit on all compiler versions from 36.0

Fixes #296.

### DIFF
--- a/delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnTypeTest.java
+++ b/delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnTypeTest.java
@@ -35,83 +35,118 @@ import org.sonar.plugins.communitydelphi.api.type.Type.StructType;
 import org.sonar.plugins.communitydelphi.api.type.TypeFactory;
 
 class IntrinsicReturnTypeTest {
-  private static final TypeFactory TYPE_FACTORY =
+  private static final TypeFactory TYPE_FACTORY_32_ALEXANDRIA =
+      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("35.0"));
+  private static final TypeFactory TYPE_FACTORY_32_ATHENS =
+      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("36.0"));
+  private static final TypeFactory TYPE_FACTORY_64_ALEXANDRIA =
       new TypeFactoryImpl(Toolchain.DCC64, CompilerVersion.fromVersionNumber("35.0"));
-  private static final TypeFactory TYPE_FACTORY_ATHENS =
+  private static final TypeFactory TYPE_FACTORY_64_ATHENS =
       new TypeFactoryImpl(Toolchain.DCC64, CompilerVersion.fromVersionNumber("36.0"));
 
   @Test
   void testLength() {
-    Type shortString = TYPE_FACTORY.getIntrinsic(IntrinsicType.SHORTSTRING);
-    Type ansiString = TYPE_FACTORY.getIntrinsic(IntrinsicType.ANSISTRING);
-    Type wideString = TYPE_FACTORY.getIntrinsic(IntrinsicType.WIDESTRING);
-    Type unicodeString = TYPE_FACTORY.getIntrinsic(IntrinsicType.UNICODESTRING);
+    Type shortString = TYPE_FACTORY_64_ALEXANDRIA.getIntrinsic(IntrinsicType.SHORTSTRING);
+    Type ansiString = TYPE_FACTORY_64_ALEXANDRIA.getIntrinsic(IntrinsicType.ANSISTRING);
+    Type wideString = TYPE_FACTORY_64_ALEXANDRIA.getIntrinsic(IntrinsicType.WIDESTRING);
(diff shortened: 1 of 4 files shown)


delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnTypeTest.java
```diff
@@ -35,83 +35,113 @@
 import org.sonar.plugins.communitydelphi.api.type.TypeFactory;
 
 class IntrinsicReturnTypeTest {
-  private static final TypeFactory TYPE_FACTORY =
+  private static final TypeFactory TYPE_FACTORY_32_RIO =
+      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("35.0"));
+  private static final TypeFactory TYPE_FACTORY_32_ATHENS =
+      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("36.0"));
+  private static final TypeFactory TYPE_FACTORY_64_RIO =
       new TypeFactoryImpl(Toolchain.DCC64, CompilerVersion.fromVersionNumber("35.0"));
```
fourls, line 43: You've jumped an extra version backwards for these variable names 😄

```suggestion
  private static final TypeFactory TYPE_FACTORY_32_ALEXANDRIA =
      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("35.0"));
  private static final TypeFactory TYPE_FACTORY_32_ATHENS =
      new TypeFactoryImpl(Toolchain.DCC32, CompilerVersion.fromVersionNumber("36.0"));
  private static final TypeFactory TYPE_FACTORY_64_ALEXANDRIA =
      new TypeFactoryImpl(Toolchain.DCC64, CompilerVersion.fromVersionNumber("35.0"));
```

fourls (approved):
Looks good! Just one note.

fourls (changes requested):
Oops, pressed the wrong button.

delphi-frontend/src/test/java/au/com/integradev/delphi/type/intrinsic/IntrinsicReturnTypeTest.java
Cirras: Oops. 👀

PR #313, opened 2024-11-13
Fix name resolution on explicitly-accessed overloaded default properties

### LINKED ISSUE #78: Default property parent overloading not handled
### Prerequisites

### SonarDelphi version

0.40.0

### SonarQube version

_No response_

### Issue description

Consider a parent with a default index property called `Items` and a child class with a default index property with the same name. When the types of the index property differ, the default property use can resolve to both parent and child declarations. 
This issue manifests in a few ways, but an example is the `Default array properties should not be explicitly referenced` check. It will raise an issue on the child's index reference and not the parent's.

### Steps to reproduce

```
type
  TParent = class(TObject)
    ...
    property Items[Index: Integer]: Integer read GetItem write SetItem; default;
  end;

  TChild = class(TParent)
    ...
    property Items[Index: String]: Integer read GetItem write SetItem; default;
  end;

begin
    TChild.Create.Items['a']; // Noncompliant
    TChild.Create.Items[1]; // Falsely compliant
end
```

_No response_

Cirras:
This PR fixes a name resolution issue where explicit references to default array properties failed to resolve to overloads from ancestor types.

Fixes #78.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Incorrect return types for `Length`, `High`, and `Low` on open/dynamic arrays depending on the
   compiler version and toolchain.
+- Name resolution failures on explicit references to default array properties with overloads on
+  ancestor types.
 
 ## [1.11.0] - 2024-11-04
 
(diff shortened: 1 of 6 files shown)


fourls (approved):
Looks good to me! Test cases make sense and everything passes.

PR #316, opened 2024-11-22
Bump `sonar-ws` and `sonar-plugin-api-impl` to 10.7.0.96327

Cirras:
This removes vulnerable protobuf-java and okio-jvm versions from our dependency tree.


fourls (approved):
If CI's happy with this, I am too.

PR #317, opened 2024-12-02
Expose URI of the analyzed file through `DelphiCheckContext::getUri`

Cirras:
This PR adds a new `DelphiCheckContext::getUri` method to expose the URI of the file under analysis.


fourls (approved):
Looks good! Tested and is accessible from custom rules plugins.

PR #319, opened 2024-12-05
Fix various intrinsic signatures around dynamic and open arrays

Cirras:
This PR fixes several intrinsic signatures to handle array size types properly.
It's a symmetrical change with #312.

Due to the current asymmetry, it's very trivial to get `PlatformDependentTruncation` false-positives with cases like:
```pas
procedure Test(Foo: TArray<Byte>; Bar: TArray<Byte>);
begin
  // Length returns NativeInt, and SetLength incorrectly expects Integer.
  SetLength(Foo, Length(Bar));
end;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Various intrinsic routines had incorrect signatures around dynamic and open arrays.
+
 ## [1.12.0] - 2024-12-02
 
 ### Added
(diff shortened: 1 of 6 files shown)


fourls (approved):
Change looks good. Just to confirm my understanding, passing an `Int64` to these intrinsics should now cause a `PlatformDependentTruncation` issue? Assuming the Delphi documentation is correct (...), this is the right behaviour.

Cirras:
> Change looks good. Just to confirm my understanding, passing an `Int64` to these intrinsics should now cause a `PlatformDependentTruncation` issue? Assuming the Delphi documentation is correct (...), this is the right behaviour.

No - and it's kind of an irritating problem.
For accuracy reasons, we return `Integer` for array sizes on 32-bit. We'd need custom handling in the `PlatformDependent*` rules to consistently detect this on 32-bit toolchains.

PR #320, opened 2024-12-05
Fix FPs around binary expressions in `PlatformDependentTruncation`

Cirras:
# Problem

Under 64-bit toolchains, `PlatformDependentTruncation` currently produces false positives around binary expressions with platform-dependent operands.

# Example

```pas
procedure Test(Nat: NativeInt);
begin
  Nat := Nat + 1;
end;
```

* On a 64-bit toolchain, the expression `Nat + 1` has result type `Int64` because the `Int64` overload of the `Add` operator intrinsic is selected.
* On a 64-bit toolchain, the assignment expression is `<NativeInt> = <Int64>`, and so the rule says "hey, `Int64` doesn't fit into `NativeInt` on 32-bit, that will truncate!"
* What the rule doesn't understand is that, on a 32-bit toolchain, a different overload of the `Add` operator intrinsic will be selected and the result type will be `Integer` - giving us an effective expression of `<32-bit NativeInt> = <Integer>` on that toolchain.

# Solution

Recursively check the operands of binary expression for platform-dependent operands - the whole expression is considered platform-dependent if any of the operands are.

## Is this the right solution?

**Kind of.** It's certainly the right solution for solving this narrowly-scoped problem with binary expressions.

In the grand scheme it's more of a half-measure that covers common cases like `NativeInt + 1` "decaying" to `Integer`/`Int64` and causing really stupid-looking false positives.

There are still holes in this rule's reasoning that we should try to tackle at some stage, but it's non-trivial and would require some solid heuristics - or support for emulating name resolution in a granular way on different toolchains from within rule implementations.

### What kind of holes?

```pas
procedure GetPlatformIndependentType(Int: Integer): Integer; overload;
begin
  Result := Int;
end;

procedure GetPlatformIndependentType(I64: Int64): Integer; overload;
begin
  Result := I64;
end;

procedure Test(Nat: NativeInt);
begin
  Nat := GetPlatformIndependentType(Nat); // This will cause false positives on 64-bit toolchains.
end;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Various intrinsic routines had incorrect signatures around dynamic and open arrays.
+- False positives around platform-dependent binary expressions in `PlatformDependentTruncation`.
 
 ## [1.12.0] - 2024-12-02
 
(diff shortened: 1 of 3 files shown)


fourls (approved):
Looks good!

PR #322, opened 2024-12-05
Include Delphi version in bug report required fields

fourls:
This PR adds a new required field "Delphi version" to the bug report issue template. This will help us triage issues that are due to various issue-specific syntaxes or behaviours.


Cirras (approved):
Looks good to me. 👍

PR #323, opened 2024-12-06
Fix incorrect type resolution around array property accesses

Cirras:
#313 introduced a regression in type resolution for plain array properties.

We started erroneously calling `addResolvedDeclaration()` before resolving array properties, where before we would exit immediately and return to argument disambiguation on the array accessor.

The effect was a subtle breakage of the type resolution around array properties. Take, for example, the expression `Foo.Bar[0]`, where `Bar` is an array property returning `string`. The old behavior would yield type `string`, while the new behavior would yield type `Char` due to the premature `addResolvedDeclaration()` call updating the `currentType` to `string` before the array accessor was processed.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
@@ -782,26 +782,31 @@ public class NameResolver {
     }
   }
 
+  private void addExplicitDefaultArrayPropertyDeclarations() {
+    if (!declarations.stream().allMatch(NameResolver::isDefaultArrayProperty)) {
+      return;
+    }
+
+    Type type = currentType;
+    if (type.isClassReference()) {
+      type = ((ClassReferenceType) type).classType();
+    } else if (type.isProcedural()) {
+      type = ((ProceduralType) type).returnType();
+    }
+
+    if (type.isStruct()) {
+      StructType structType = (StructType) TypeUtils.findBaseType(type);
+      NameDeclaration declaration = Iterables.getLast(declarations);
+      ((StructTypeImpl) structType)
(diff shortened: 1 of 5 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
```diff
@@ -782,9 +782,17 @@ private void handleParenthesizedExpression(ParenthesizedExpressionNode parenthes
     }
   }
 
-  private boolean handleDefaultArrayProperties(ArrayAccessorNode accessor) {
-    if (!declarations.isEmpty()
-        && declarations.stream().allMatch(NameResolver::isDefaultArrayProperty)) {
+  private boolean handleExplicitArrayProperties() {
```
fourls, line 785: This method is quite confusing for a few reasons:

- It does multiple things (determines if the name is an explicit array property, and also adds declarations if the name is a default array property)
- It doesn't handle explicit array properties, it exits early
- Its handling logic is for default array properties

I think it would probably be suitable to inline the early exit for explicit array properties back into `handleDefaultArrayProperties`, rename this routine to `addDefaultArrayPropertyDeclarations` or similar, and call it after the early exit, immediately before `addResolvedDeclaration`. Thoughts?

fourls (changes requested):
Logic of the change looks good - the extra tests are a really good idea. Just one thought about the method layout.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
Cirras: Makes sense to me.

fourls (approved):
Looks good.

PR #324, opened 2024-12-10
Implement Dependency Submission workflow

Cirras:
@spiltcoffee, a replacement for the automatic dependency submission.

The automatic dependency submission always fails on release commits because the new versions for `delphi-tokens-generator-maven-plugin` and `license-maven-plugin` are unavailable until they've been published.

This new workflow installs everything locally before running the dependency submission.

### DIFF
--- a/.github/workflows/scan-dependencies.yml
+++ b/.github/workflows/scan-dependencies.yml
@@ -1,0 +1,27 @@
+name: Scan Dependencies
+
+on:
+  push:
+    paths:
+      - '**/pom.xml'
+    branches:
+      - 'master'
+
+permissions:
+  contents: write
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: maven
+      - name: Maven Install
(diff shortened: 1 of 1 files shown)


.github/workflows/scan-dependencies.yml
```diff
@@ -0,0 +1,29 @@
...
+permissions:
+  contents: write
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
spiltcoffee, line 19: Is all history needed for this check? Wouldn't the default of `fetch-depth: 1` suffice for this check?

spiltcoffee (changes requested):
One question, otherwise looking good

.github/workflows/scan-dependencies.yml
spiltcoffee: Based on https://github.com/actions/checkout
```yml
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

Cirras: Oops, this was just copied from the build workflow - where we want fetch-depth 0 for sonar SCM analysis.

spiltcoffee (approved):
lgtm :+1:

PR #331, opened 2025-01-20
Handle empty anonymous methods in `EmptyRoutine` instead of `EmptyBlock`

### LINKED ISSUE #315: EmptyBlock raises issue on empty anonymous methods, but EmptyRoutine does not
### Prerequisites

### SonarDelphi version

1.11.0

### SonarQube version

_No response_

### Issue description

EmptyBlock currently raises an issue on an empty anonymous method, while EmptyRoutine (which should) does not.

For example, the following raises a single EmptyBlock issue:

```
MyProcedure := procedure begin
    // I am empty for a reason!
  end;
```

This should be ignored by EmptyBlock, since the description states:

> **Exceptions**
> * Empty routine bodies (covered by the EmptyRoutine rule)

EmptyRoutine, on the other hand, should pick this up.

### Steps to reproduce

Scan the following code with SonarDelphi:

```delphi
unit Test;

interface

implementation

uses System.SysUtils;

function GetDummyProc: TProc;
begin
  Result := procedure begin
      // I am empty for a reason
    end;
end;

end.
```

_No response_

Cirras:
This PR excludes anonymous methods from `EmptyBlock` and flags them in `EmptyRoutine` instead.

Fixes #315.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Library path (`DelphiLibraryPath`/`DelphiTranslatedLibraryPath`)
   - Browsing path (`DelphiBrowsingPath`)
   - Standard library
+- Empty anonymous methods are now flagged in `EmptyRoutine`.
+- **API:** `AnonymousMethodNode::getStatementBlock` method.
+- **API:** `AnonymousMethodNode::isEmpty` method.
 
 ## [1.12.2] - 2025-01-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Library path (`DelphiLibraryPath`/`DelphiTranslatedLibraryPath`)
   - Browsing path (`DelphiBrowsingPath`)
   - Standard library
+- Empty anonymous methods are now ignored in `EmptyBlock`.
 - Empty anonymous methods are now flagged in `EmptyRoutine`.
 - **API:** `AnonymousMethodNode::getStatementBlock` method.
 - **API:** `AnonymousMethodNode::isEmpty` method.
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/EmptyRoutineCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/EmptyRoutineCheck.java
@@ -20,10 +20,12 @@ package au.com.integradev.delphi.checks;
 
 import au.com.integradev.delphi.utils.InterfaceUtils;
 import org.sonar.check.Rule;
+import org.sonar.plugins.communitydelphi.api.ast.AnonymousMethodNode;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.RoutineImplementationNode;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
 import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.check.FilePosition;
 import org.sonar.plugins.communitydelphi.api.symbol.declaration.RoutineDirective;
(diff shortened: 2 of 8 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/EmptyRoutineCheck.java
```diff
@@ -42,6 +44,18 @@ public DelphiCheckContext visit(RoutineImplementationNode routine, DelphiCheckCo
     return super.visit(routine, context);
   }
 
+  @Override
+  public DelphiCheckContext visit(AnonymousMethodNode anonymousMethod, DelphiCheckContext context) {
+    if (anonymousMethod.isEmpty() && shouldAddViolation(anonymousMethod)) {
```
fourls, line 49: It seems odd to have `anonymousMethod.isEmpty()`, one of the two criteria for whether a violation should be added, outside of `shouldAddViolation`. Could we refactor so that this is inside `shouldAddViolation`? Same with the existing `RoutineImplementationNode` code.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Library path (`DelphiLibraryPath`/`DelphiTranslatedLibraryPath`)
   - Browsing path (`DelphiBrowsingPath`)
   - Standard library
+- Empty anonymous methods are now ignored in `EmptyBlock`.
+- Empty anonymous methods are now flagged in `EmptyRoutine`.
```
fourls, line 22: API additions to `AnonymousMethodNode` should be mentioned in the changelog.

delphi-frontend/src/main/java/org/sonar/plugins/communitydelphi/api/ast/AnonymousMethodNode.java
```diff
@@ -41,4 +41,8 @@ public interface AnonymousMethodNode extends ExpressionNode {
   boolean isFunction();
 
   boolean isProcedure();
+
+  CompoundStatementNode getStatementBlock();
+
+  boolean isEmpty();
```
fourls, line 47: What's your reasoning for having `isEmpty()` as a separate function on `AnonymousMethodNode`, when an API consumer could just use the more declarative `getStatementBlock().isEmpty()`?

fourls (changes requested):
Looks pretty good! I have a couple of comments.

delphi-frontend/src/main/java/org/sonar/plugins/communitydelphi/api/ast/AnonymousMethodNode.java
Cirras: Symmetry with the `RoutineNode` API, mostly.

CHANGELOG.md
Cirras: Good catch, knew I was forgetting something!

delphi-checks/src/main/java/au/com/integradev/delphi/checks/EmptyRoutineCheck.java
Cirras: Done.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/EmptyRoutineCheck.java
```diff
@@ -59,4 +76,11 @@ private static boolean shouldAddViolation(RoutineImplementationNode routine) {
         && !declaration.hasDirective(RoutineDirective.VIRTUAL)
         && !InterfaceUtils.implementsMethodOnInterface(declaration);
   }
+
+  private static boolean shouldAddViolation(AnonymousMethodNode anonymousMethod) {
+    if (!anonymousMethod.isEmpty()) {
+      return false;
+    }
+    return anonymousMethod.getStatementBlock().getComments().isEmpty();
```
fourls, line 84: This can be a single statement:
```suggestion
  return anonymousMethod.isEmpty() && anonymousMethod.getStatementBlock().getComments().isEmpty();
```

fourls (changes requested):
Just one note. Looks good otherwise.

fourls (approved):
Looks good!

PR #333, opened 2025-02-04
Tidy the `AnonymousMethodNode` API

Cirras:
This PR fixes a bug in the `getStatementBlock` implementation, and adds a `getDeclarationSection` method.


fourls (approved):
Looks good. On a grammar level it is theoretically possible for `getStatementBlock()` to be nullable, but it's not valid Delphi so it doesn't bother me.

PR #334, opened 2025-02-04
Deprioritize `sonar.sources` in the analysis search path

Cirras:
The previous approach (where `sonar.sources` took first priority in the analysis search path) caused some surprising/undesirable behavior in projects where there are multiple units with the same name.

It becomes very hard in that scenario for a user to configure the scan in a way where the correct unit will win during import resolution, even if the search path was correct in the dproj file.


fourls (approved):
Looks good!

PR #335, opened 2025-02-10
Fix parsing ambiguity around adjacent `>` and `=` tokens

### LINKED ISSUE #318: parsing error on method parameter declaration "cAdditionalSQL: IDictionary<string, string>=nil"
### Prerequisites

### SonarDelphi version

1.11.0

### SonarQube version

10.7

### Issue description

We had this error during parsing:

> ERROR Error while analyzing D:\test\ant\test.pas au.com.integradev.delphi.file.DelphiFile$DelphiFileConstructionException: Failed to construct DelphiFile (line 7:21 no viable alternative at input '(')

adding spaces before the = sign  in the parameter declaration will solve the problem

### Steps to reproduce

try analyze the following code.
add spaces before and after the = sign (in the class Public methods declaration) to remove the error

### Case 1

```
unit test;
Interface
Type
  TAnagrafica = Class(TObject)
   Public
    procedure GetSlaveItemsBad(cAdditionalSQL: IDictionary<string, string>=nil); // bad declaration --> generate parsing error
    procedure GetSlaveItemsGood(cAdditionalSQL: IDictionary<string, string> = nil); // good declaration --> no error  
  End;
Implementation
End.
```

### another code example
In this case a 'missing FORWARD' message error is produced

```
unit testassignment;
interface
type
 myclass = Class(Tobject)
   Public
	Procedure myMethod;
 End;
 
Implementation


 Procedure myClass.myMethod;
 Var
    B : Boolean;
 const
  aEntityTableNames: tArray<string>=[EntityField_TableName,
                                    EntityAction_TableName,
                                    Linguaggio_TableName,
                                    Postazione_TableName,
                                    Config_EntityName,
                                    Utente_EntityName,
                                    Gruppo_TableName];	

Begin
End;
  
 end.
```

Cirras:
This PR fixes a grammar ambiguity around adjacent `>` and `=` tokens. (Fixes #318)

We now create `>=` tokens during parsing by combining the `>`/`=` tokens, instead of combining them at the lexing stage.
(For consistently, we're doing this for "less than equal" tokens as well)

This allows us to only create `>=` tokens when we're parsing a binary expression.

As a result, we can now successfully parse code that looks like this:
```pas
const Foo: TArray<Byte>=[1, 2, 3];
```
Previously this would create a `>=` token in the middle and make everything fall over.

### DIFF
--- a/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
+++ b/delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
@@ -71,6 +71,8 @@ tokens {
   TkArgument;
   TkAnonymousMethod;
   TkAnonymousMethodHeading;
+  TkLessThanEqual;
+  TkGreaterThanEqual;
 }
 
 @header
@@ -328,11 +330,17 @@ import org.apache.commons.lang3.StringUtils;
     return t;
   }
 
-  private Token combineLastNTokens(int count) {
+  private Token combineLastNTokens(int type, int count) {
     CommonToken firstToken = (CommonToken) input.LT(-count);
     CommonToken lastToken = (CommonToken) input.LT(-1);
+    lastToken.setType(type);
     lastToken.setStartIndex(firstToken.getStartIndex());
     return lastToken;
+  }
+
+  private BinaryExpressionNodeImpl createBinaryExpression(Object operator) {
+    Token token = adaptor.getToken(operator);
+    return new BinaryExpressionNodeImpl(token);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -328,13 +330,19 @@ import org.apache.commons.lang3.StringUtils;
...
     CommonToken lastToken = (CommonToken) input.LT(-1);
+    lastToken.setType(type);
     lastToken.setStartIndex(firstToken.getStartIndex());
     return lastToken;
   }
 
+  private BinaryExpressionNodeImpl createBinaryExpression(Object operator) {
+    Token token = adaptor.getToken(operator);
+    return new BinaryExpressionNodeImpl(token);
+  }
```
fourls, line 344: To make this a little more general-purpose, could we make this generic? Maybe something like `transformNode<T>`?

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -1447,7 +1457,7 @@ Alpha                   : 'a'..'z'
                         // surrogate pairs
                         | ('\uD800'..'\uDBFF') ('\uDC00'..'\uDFFF')
                         ;
-fragment
+  fragment
```
fourls, line 1460: What's the reason for this indent?

fourls (changes requested):
Looks good! A couple questions.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
Cirras: Oops, apparently didn't push the fix for this (or the comments for the weird new parser rules)

Cirras: That's fixed now.

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
Cirras: No, not really.

You'd have to pass a `Class<T>` into the function and construct the object via reflection, which I imagine would be a quite a bit slower than what we're currently doing.

We probably don't need a general-purpose version anyway, as I'd rather eat a bowl of glass (without any milk) than pepper the grammar with something like `transformNode`. It's a heinous band-aid.

fourls: Ugh, Java generics.

fourls (approved):
Looks good!

PR #336, opened 2025-02-14
Use standard functional interfaces over Guava

Cirras:
This PR replaces 2 usages of the Guava `Supplier` type with the standard `java.util.function.Supplier`.


jgardn3r (approved):
Deceptively simple.

PR #337, opened 2025-02-14
Improve coverage in `RedundantJumpCheck`

Cirras:
This PR simplifies some code and adds some tests to improve coverage in the new `RedundantJumpCheck`.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantJumpCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantJumpCheck.java
@@ -23,12 +23,10 @@ import au.com.integradev.delphi.cfg.ControlFlowGraphFactory;
 import au.com.integradev.delphi.cfg.api.Block;
 import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
 import au.com.integradev.delphi.cfg.api.Finally;
-import au.com.integradev.delphi.cfg.api.Linear;
 import au.com.integradev.delphi.cfg.api.UnconditionalJump;
 import org.sonar.check.Rule;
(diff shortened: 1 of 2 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantJumpCheck.java
```diff
@@ -69,19 +63,20 @@ private void checkBlock(Block block, DelphiCheckContext context) {
...
+      return;
+    }
+
     if (!isContinueOrExit(routineNameDeclaration)
         || isExitWithExpression(routineNameDeclaration, terminator)) {
       return;
     }
 
     Block successor = jump.getSuccessor();
-    successorWithoutJump = nonEmptySuccessor(successorWithoutJump);
```
fourls, line 82: I'm a little nervous about this assumption that `UnconditionalJump` can never be followed by a `Linear` block. `UnconditionalJump` occurs many times in a CFG, most notably for every loop statement. Are we certain that this case can never come up?

fourls (commented):
Looks good, although I have a question about the `RedundantJumpCheck` change.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/RedundantJumpCheck.java
Cirras: @jgardn3r and I looked over the CFG and concluded that the `nonEmptySuccessor` we're looking for here is basically mythical. See `ControlFlowGraphImpl::prune`, where empty linear blocks are stripped from the graph.

fourls: Sure, if it's cosigned by Josh 🫡

fourls (approved):

PR #339, opened 2025-02-26
Fix IndexLastListElement issues on assignments and var/out

fourls:
This PR updates IndexLastListElement to ignore the left-hand side of expression statements. This is a current false-positive, since `TList.Last` is a function and is not eligible to be assigned to.


fourls:
Oops! Forgot the changelog.

PR #341, opened 2025-03-04
Fix file pointer errors around `>=` and `<=` tokens

Cirras:
This PR fixes instances of  `IllegalArgumentException` being raised by the CPD executor on lines that end with `>=` or `<=`,, caused by file pointer errors.

`>=` and `<=` token positions were incorrect after token combination had occurred.
The solution is to adjust the token `line` and `charPositionInLine` as part of the combination step, where we previously only adjusted the token start/stop indices.


fourls (approved):
Looks good!

PR #342, opened 2025-03-05
Fix "tokens out of order" CPD errors around `>=` and `<=` tokens

Cirras:
Token combination now creates a new token instead of mutating the last token in the group of combined tokens.

This was causing problems because the combined token would be reflected back to the original list of "raw tokens" returned by `DelphiFile::getTokens`, resulting in overlapping tokens being reported to the sonar API.

The mutability of the raw tokens is a point of brittleness that we could address, but it would have significant performance implications because it would involve constructing a copy of every token in every file.


fourls:
Stretching my terminal window to its limits...
image

fourls (approved):
Looks good! Confirmed that this fixes the errors and that a test scan looks back to normal.

PR #348, opened 2025-03-17
Exclude types annotated with attributes in `UnusedType`

Cirras:
This PR adds an exclusion to `UnusedType` for types that are annotated with custom attributes.

It's unclear why this wasn't already the case, as the same exclusion is already present in `UnusedRoutine`, `UnusedProperty`, and `UnusedField`.


fourls (approved):
Looks good! This will definitely improve the experience for RTTI-heavy projects.

PR #351, opened 2025-03-27
Define the `WEAK_NATIVEINT` conditional on Delphi 12+

Cirras:
This PR adds support for the undocumented `WEAK_NATIVEINT` predefined conditional symbol.
It's defined from Delphi 12 onward.

This was discovered due to a strange false positive on Delphi 12:
image

It turns out that the Delphi 12 RTL still declares `TNativeIntHelper` and gates it behind this `WEAK_NATIVEINT` symbol.
Without the symbol defined, the analyzer would see the `TNativeIntHelper` record helper defined and overriding `TIntegerHelper` since (as weak aliases) they are the same type now.

This caused the analyzer to think that `TNativeIntHelper.Parse` was being called, which (of course) returns a `NativeInt` and confuses the `PlatformDependentTruncation` check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for the `WEAK_NATIVEINT` symbol, which is defined from Delphi 12 onward.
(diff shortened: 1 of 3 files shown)


fourls (approved):
Looks good!

fourls (approved):

PR #352, opened 2025-03-27
Add support for undocumented `varargs` intrinsics

Cirras:
This PR adds support for 4 undocumented intrinsics that are usable within `varargs` routines.

Signatures for these intrinsics can be found in `System`:
```pas
  // procedure VarArgStart(var ArgList: TVarArgList);
  // function  VarArgGetValue(var ArgList: TVarArgList; ArgType: Type): ArgType;
  // procedure VarArgCopy(var DestArgList, SrcArgList: TVarArgList);
  // procedure VarArgEnd(var ArgList: TVarArgList);
```

We can't easily inject intrinsics with specific concrete types like `System.TVarArgList`, so I've opted for the "good enough" solution of using untyped parameters for these. I can't see any obvious way for this to result in a less inaccurate analysis (on source code that successfully compiles).

### See also
* [How to create a Delphi variadic method similar to Write/Writeln without requiring brackets for arguments?](https://en.delphipraxis.net/topic/12789-how-to-create-a-delphi-variadic-method-similar-to-writewriteln-without-requiring-brackets-for-arguments/?do=findComment&comment=99558)
* [\[fpc-devel\] C-block reference syntax (blocker for 3.2)](https://lists.freepascal.org/fpc-devel/2019-December/042285.html)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for the `WEAK_NATIVEINT` symbol, which is defined from Delphi 12 onward.
+- Support for undocumented intrinsics usable within `varargs` routines:
+  - `VarArgStart`
+  - `VarArgGetValue`
+  - `VarArgCopy`
+  - `VarArgEnd`
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


fourls (approved):
Looks good! Now I can happily scan all my most upsetting Win64 variadics!!

PR #353, opened 2025-03-31
Implement `ParsingError` check

### LINKED ISSUE #301: Parsing errors should be better handled
### Prerequisites

- [X] This question has not already been answered.
- [X] This question is about SonarDelphi, not SonarQube or Delphi.

### Inquiry

Hello!

In my company, we have hundreds of Delphi projects wich we run SonarQube analysis (with Sonar Delphi). In the analysis there are a bunch of parsing errors, mostly related to encoding problems. 

Checking some other language plugins, I noticed that almost all of them have a "Parsing error" rule , wich raises an issue when a file fails to be parsed, rather then raise an error log on SonarScanner execution.

image

image

So, I wanted to start the discussion asking if an optional rule with this objective makes sense, raising issues when a file fails to be parsed if the rule is active, or keep raising an error log if it's not.

Cirras:
This PR implements a check to flag files where parsing failures have occurred.
The official analyzers from SonarSource have similar rules. Like them, I've opted not to put it in the default quality profile.

Closes #301.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `ParsingError` analysis rule, which flags files where parsing failures occurred.
 - Support for the `WEAK_NATIVEINT` symbol, which is defined from Delphi 12 onward.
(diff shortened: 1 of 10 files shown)


fourls (approved):
Looks good!

PR #354, opened 2025-03-31
Fix overload resolution failures on the variant overload of `Copy`

Cirras:
This PR fixes an overload resolution issue involving the `Copy` intrinsic.
We didn't previously register the variant overload `Copy(Variant, Integer, Integer)`.

This caused overload resolution failures when calling `Copy` on a variant value.


fourls (approved):
Looks good!

PR #355, opened 2025-03-31
Handle inline `var` and `const` in `PlatformDependentTruncation`

Cirras:
This PR fixes false negatives in `PlatformDependentTruncation` around inline `var` and `const` assignments.


fourls (approved):
Looks good!

PR #357, opened 2025-04-04
Add support for `CompilerVersion` in compiler directive expressions

Cirras:
This PR adds support for the `CompilerVersion` constant in compiler directive expressions.

It also improves the evaluation of:
- `System.True`
- `System.False`
- `System.Defined`
- `System.SizeOf`

This is more of a bolt-on hack than anything. It's worth doing for now because `CompilerVersion` is frequently used in directive expressions and can be statically known from configuration (`sonar.delphi.compilerVersion`).

For more accurate evaluation of compiler directive expressions, we're still blocked on #261.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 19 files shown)


fourls (approved):
Looks good.

PR #371, opened 2025-04-22
Add check for ambiguous bitwise not operations

### LINKED ISSUE #97: New rule: Set containment checks should be enclosed in parentheses
### Prerequisites

### Suggested rule title

Set containment checks should be enclosed in parentheses

### Rule description

In Delphi, the `not` unary operator is stronger than the `in` binary operator, meaning that `not MyVar in MySet` is parsed as `(not MyVar) in MySet` instead of the generally expected `not (MyVar in MySet)`. 

This can be particularly confusing for integers, for which `not` is a perfectly valid operator - the example below incorrectly prints `MyByte is neither 3 or 252`.

```delphi
procedure RunTest;
var
  MyByte: Byte;
begin
  MyByte := 3;
  Assert(not MyByte = 252);

  if not MyByte in [3, 252] then begin
    Writeln('MyByte is neither 3 or 252');
  end
  else begin
    Writeln('MyByte is 3 or 252!');
  end;
end;
```

This rule would pick up these cases and suggest using parentheses to make the order of operations clear and intentional.

### Rationale

This is a case in which the compiler's interpretation is almost never what the programmer intended. Even if the programmer did mean to check if `not X` was in the array, parentheses are needed to make this clear to other maintainers.

fourls:
Fixes #97. This rule works by highlighting any `in` binary expression with one operand being a bitwise `not`. ~I haven't added it to the default quality profile~ I have now added it to the default quality profile, since it is a rare and admittedly confusing case.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
@@ -1,0 +1,67 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2025 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.communitydelphi.api.ast.BinaryExpressionNode;
+import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
+import org.sonar.plugins.communitydelphi.api.ast.UnaryExpressionNode;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.check.FilePosition;
+import org.sonar.plugins.communitydelphi.api.operator.BinaryOperator;
+import org.sonar.plugins.communitydelphi.api.operator.UnaryOperator;
+import org.sonar.plugins.communitydelphi.api.reporting.QuickFix;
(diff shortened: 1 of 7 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+    UnaryExpressionNode unaryNode = (UnaryExpressionNode) node;
+
+    if (isBitwiseNot(unaryNode)) {
+      context
+          .newIssue()
+          .onFilePosition(FilePosition.from(unaryNode.getToken()))
+          .withMessage("Parenthesize this bitwise NOT operation.")
+          .withQuickFixes(
+              QuickFix.newFix("Parenthesize bitwise NOT")
```
Cirras, line 57: nit: do we like the screaming NOT or a quoted 'not' better for these messages?

Cirras (changes requested):

delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
fourls: I personally think uppercase is better but I'm not opinionated.

Cirras: image

fourls: Updated with quoted note.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/ExplicitBitwiseNot.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "Ambiguous bitwise NOT operations should be parenthesized",
```
Cirras, line 2: Missed one.
```suggestion
  "title": "Ambiguous bitwise 'not' operations should be parenthesized",
```

delphi-checks/src/test/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheckTest.java
```diff
@@ -0,0 +1,120 @@
...
+    CheckVerifier.newVerifier()
+        .withCheck(new ExplicitBitwiseNotCheck())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendImpl("function Test: Integer;")
+                .appendImpl("begin")
+                .appendImpl("  Result := not UnresolvedIdentifier in [2];")
+                .appendImpl("end;"))
+        .verifyNoIssues();
+  }
```
Cirras, line 119: Can we add a test for `not Func in [2]`?

Cirras (changes requested):

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/ExplicitBitwiseNot.json
fourls: Fixed

PR #372, opened 2025-04-23
Add check for array iterations where the high bound is out of range

### LINKED ISSUE #95: New rule: Array iteration should be in range
### Prerequisites

### Suggested rule title

Array iteration should be in range

### Rule description

This rule would pick up on cases where a loop iterator may become out of range of the underlying array, by determining that the upper bound of iteration is less than or equal to the length of the array - 1. For example:

```delphi
for I := 0 to Length(MyArr) do begin
  // ...
end;
```

This could also be extended to lists:

```delphi
for I := 0 to MyList.Count do begin
  // ...
end;
```

### Rationale

It's a very common problem to write `for I := 0 to Length(MyArr)`, as it looks correct at first glance. This rule would catch those cases before causing a range check error.

fourls:
Fixes #95. I haven't added this to the Sonar way quality profile, but it probably could be added depending on how noisy it is. I'll run some tests and see.

### DIFF
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/IterationPastHighBoundCheckTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/IterationPastHighBoundCheckTest.java
@@ -1,0 +1,344 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2025 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.checks;
+
+import au.com.integradev.delphi.builders.DelphiTestUnitBuilder;
+import au.com.integradev.delphi.checks.verifier.CheckVerifier;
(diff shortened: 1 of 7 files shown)


Cirras:
I'd definitely lean towards the default quality profile since this is a common bug pattern.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/IterationPastHighBoundCheckTest.java
```diff
@@ -0,0 +1,305 @@
...
+  void testIterationToLengthMinusOneShouldNotAddIssue() {
+    CheckVerifier.newVerifier()
+        .withCheck(new IterationPastHighBoundCheck())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendImpl("procedure Test(MyArr: TArray<string>);")
+                .appendImpl("var")
+                .appendImpl("  I: Integer;")
+                .appendImpl("begin")
+                .appendImpl("  for I := 0 to Length(MyArr) - 1 do begin")
```
CoppensBart, line 36: Can we also check if using `Pred()` [https://docwiki.embarcadero.com/Libraries/Athens/en/System.Pred] instead of `-1` works?
`.appendImpl(" for I := 0 to Pred(Length(MyArr)) do begin")`

fourls: Because of the way the rule works, this test isn't asserting that the `-1` is detected as such - it's detecting that the expression is not of the form `Length(...)`. So something like `Pred(Length(...))` is already correctly ignored.

That being said, I'll add the test - could be a useful addition. Thanks!

CoppensBart: Ok, thanks for the clarification

fourls:
- Added to the default quality profile, and
- Added a test that `Pred(X.Count)` does not add an issue

Cirras (changes requested):
In the interest of leaning away from false positives, I think we should make this rule more conservative and only flag cases where low = 0 and high = count.

We can then iteratively make it smarter and broader if we see ways to do so without introducing noise/FPs.

fourls:
Updated to only flag cases where low = 0.

Cirras (approved):
Looks good to me. 👍

PR #375, opened 2025-05-05
Upgrade google-java-format to 1.26.0

Cirras:
This also bumps our minimum JDK requirement to 17, as it's a requirement starting from google-java-format 1.25.0.


fourls (approved):
Looks good!

PR #366, opened 2025-04-17
Disable integration tests until SonarSource fixes the orchestrator

Cirras:
See: https://community.sonarsource.com/t/fail-to-load-versions-of-org-sonarsource-sonarqube-sonar-application/138865

PR #378, opened 2025-05-07
Upgrade orchestrator to 5.6.1.2597 and restore integration tests

Cirras:
The latest version of orchestrator fixes a bug where maven metadata couldn't be parsed, which was failing our integration tests before. (see #366)


fourls (approved):
Looks good.

PR #379, opened 2025-05-07
Tidy up `ExpressionEvaluator`

Cirras:
This PR does some minor QA in the recently-added `ExpressionEvaluator`.

I've also enabled the "Control structures should use curly braces" rule on SonarCloud. It shouldn't raise any issues, as the rest of the codebase consistently uses curly braces.


fourls (approved):
More newlines = more quality.

PR #381, opened 2025-05-07
Disambiguate index property stored methods correctly

fourls:
When a property has an `index` specifier, its `stored` specifier must be a function with a single `Integer` argument.


fourls:
I've also fixed the duplicate processing of the `stored` directive, but not the others since that's covered by #377.

Cirras:
> I've also fixed the duplicate processing of the stored expression, but not the others since that's covered by https://github.com/integrated-application-development/sonar-delphi/pull/377.

This actually failed the `StoredProperties` test (featuring a `stored` specifier backed by a constant) because the top-level `PropertyNode` visit calls `disambiguateParameters` and erroneously eliminates non-invocable declarations.

I've restored the double-visit by the `SymbolTableVisitor`.

I'll rebase #377 over these changes and fix the double-visit of `stored` in the same commit `read` and `write`.

Cirras (approved):

PR #382, opened 2025-05-08
Fix various name resolution errors around routine references

Cirras:
This PR improves name resolution on routine references in:
- assignment to inline `var`/`const` declarations
- assignment to standard `var` declarations
- assignment to standard typed `const` declarations
- constant arrays
  - This required improvements to type resolution for constant array expressions.
    Multidimensional constant arrays are especially improved, as their subarrays couldn't **ever** resolve properly before.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
@@ -38,11 +38,14 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.sonar.plugins.communitydelphi.api.ast.AnonymousMethodNode;
+import org.sonar.plugins.communitydelphi.api.ast.ArrayExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.ArrayIndicesNode;
 import org.sonar.plugins.communitydelphi.api.ast.ArrayTypeNode;
 import org.sonar.plugins.communitydelphi.api.ast.AssignmentStatementNode;
 import org.sonar.plugins.communitydelphi.api.ast.AttributeListNode;
 import org.sonar.plugins.communitydelphi.api.ast.AttributeNode;
+import org.sonar.plugins.communitydelphi.api.ast.ConstDeclarationNode;
+import org.sonar.plugins.communitydelphi.api.ast.ConstStatementNode;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.ForInStatementNode;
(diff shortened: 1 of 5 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
```diff
@@ -517,18 +523,18 @@ private static boolean handleRoutineReference(
...
-      }
+      return expression != assignee && handleRoutineReference(assignee.getType(), resolver);
+    } else if (parent instanceof VarDeclarationNode) {
+      return handleRoutineReference(((VarDeclarationNode) parent).getType(), resolver);
+    } else if (parent instanceof ConstDeclarationNode) {
+      return handleRoutineReference(((ConstDeclarationNode) parent).getTypeNode(), resolver);
+    } else if (parent instanceof VarStatementNode) {
+      return handleRoutineReference(((VarStatementNode) parent).getTypeNode(), resolver);
+    } else if (parent instanceof ConstStatementNode) {
+      return handleRoutineReference(((ConstStatementNode) parent).getTypeNode(), resolver);
```
fourls, line 534: The type node is potentially null for `ConstDeclarationNode`, `VarStatementNode`, and `ConstStatementNode`. This just looks like it will fail to resolve any cases where the type is inferred?

delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/RoutineReferenceAssignedToProcedural.pas
fourls, line 1: Can we add cases to this file where the const/var is not explicitly typed?

fourls (changes requested):
I'm a little confused about how this code will resolve declarations where the type is inferred rather than notated.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
Cirras: That's because it will - you can't assign routine references unless the procedural type is explicitly declared.
When this method returns false, the name resolver falls through to the standard primary expression resolution and tries to continue resolving as if its an invocation with an implicit empty argument list.

delphi-frontend/src/test/resources/au/com/integradev/delphi/symbol/RoutineReferenceAssignedToProcedural.pas
Cirras: You can't assign routine references unless the procedural type is explicitly declared.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
fourls: > you can't assign routine references unless the procedural type is explicitly declared.

Can we add a comment here to that effect? I don't think it's obvious in this code that that is intentional.

fourls (changes requested):
Looks good, I just think we should add a comment to this code - since there's a little bit of context that's unclear from just reading.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolutionHelper.java
Cirras: Sure.

fourls (approved):
Looks good!

PR #383, opened 2025-05-09
Improve handling for various intrinsics

Cirras:
This PR does a few things.

Improves the way we handle intrinsic routines by:
- deferring their injection into `System` until the end of the `interface` section
- resolving references to public `System` types in their parameter lists and return types

Adds support for:
- `AtomicCmpExchange128`
- `GetTypeKind`
- `OpenString` type

Improves signatures of:
  - `IsManagedType`
  - `VarArgStart`
  - `VarArgGetValue`
  - `VarArgCopy`
  - `VarArgEnd`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for the `AtomicCmpExchange128` intrinsic routine.
+- Support for the `GetTypeKind` intrinsic routine.
 - **API:** `TypeParameterNode::getTypeParameters` method.
 - **API:** `InterfaceTypeNode::getGuidExpression` method.
 - **API:** `AttributeNode::getExpression` method.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Issue locations no longer span the entire routine declaration in `RoutineName`.
+- Improve type modeling around the `VarArg*` intrinsic routines.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for the `AtomicCmpExchange128` intrinsic routine.
 - **API:** `TypeParameterNode::getTypeParameters` method.
 - **API:** `InterfaceTypeNode::getGuidExpression` method.
 - **API:** `AttributeNode::getExpression` method.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Type resolution failures on `as` casts where the type is returned from a routine invocation.
 - Inaccurate type resolution when calling a constructor on a class reference type.
 - Grammar ambiguity causing attributes to be misinterpreted as interface GUIDs.
+- Failure to resolve invocations of `System.IsManagedType` where a value is passed.
 
 ## [1.16.0] - 2025-05-09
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Support for `GetTypeKind` intrinsic.
+- Support for `OpenString` intrinsic type.
+
+### Changed
+
+- Improve `VarArg*` intrinsic signatures and make them stricter.
+
+### Fixed
+
+- The `IsManagedType` intrinsic would only accept type references - it now accepts values.
```
fourls, line 22: This is phrased a little oddly - a user reading this probably wouldn't understand that this just means that `IsManagedType` invocations now correctly resolve when passed a value. Maybe

```suggestion
- Failure to resolve invocations of `System.IsManagedType` where a value is passed.
```

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Support for `AtomicCmpExchange128` intrinsic.
+- Support for `GetTypeKind` intrinsic.
+- Support for `OpenString` intrinsic type.
+
+### Changed
+
+- Improve `VarArg*` intrinsic signatures and make them stricter.
```
fourls, line 18: This is pretty vague and also probably fits into Fixed rather than Changed. 
Maybe

```suggestion
- Improve type modelling around the `VarArg*` intrinsic routines.
```

fourls (changes requested):
Code change looks good. I have a few nitpicks :)

CHANGELOG.md
Cirras: Sure.
It should probably still be in `Changes` - it's a pure QA/accuracy change, since there was no way for the previous modeling to result in invocation resolution failures.

delphi-frontend/src/test/java/au/com/integradev/delphi/executor/DelphiSymbolTableExecutorTest.java
```diff
@@ -597,6 +597,20 @@ void testDefaultIntrinsic() {
     verifyUsages(22, 10, reference(32, 2));
   }
 
+  @Test
+  void testVarArgIntrinsics() {
+    execute("intrinsics/VarArgIntrinsics.pas");
+    verifyUsages(7, 10, reference(21, 4));
+    verifyUsages(14, 2, reference(17, 14), reference(18, 13), reference(21, 24), reference(24, 12));
+    verifyUsages(15, 2, reference(18, 21));
```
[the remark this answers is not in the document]
Cirras: No - which kind of sucks.
Do you think we should remove them?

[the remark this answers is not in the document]
fourls: I think so, just because it gives a false sense of security if we appear to be testing this stuff. Ideally we could find an alternative way to test this, but none are coming to mind.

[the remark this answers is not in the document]
Cirras: Fair enough - I'll remove these assertions.

CHANGELOG.md
```diff
@@ -9,13 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support for `AtomicCmpExchange128` intrinsic.
+- Support for `GetTypeKind` intrinsic.
```
fourls, line 13: Should these be "intrinsic routine"? It looks a little odd having both "intrinsic" and "intrinsic type".

fourls (changes requested):

CHANGELOG.md
Cirras: Hm. Referring to the routines as just "intrinsics" is consistent with other changelog entries. I agree that it's (a little) confusing though since there's a type in the mix this time. 
On balance I like the current wording, but I'd be happy to add "routine" if you want.

fourls: The changelog entries are inconsistent and vary between `intrinsic`, `intrinsic routine`, and `intrinsic method`. I'm not particularly worried about that. Thoughts on `Support for the <x> intrinsic` rather than `Support for <x> intrinsic`? The former sounds better to me.

PR #384, opened 2025-05-13
Fix various type resolution failures

fourls:
This PR fixes:

- Name resolution failures around generic routine invocations, mostly because we were not specializing type constraints
- Type resolution failures on `as` casts where the type expression is a function (e.g. `GetFooClass`)
- Type resolution failures on constructors for class reference types (e.g. `TFooClass.Create`, which creates a `TFoo`)

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
@@ -1103,17 +1103,27 @@ public class NameResolver {
   }
 
   private void resolveReturnType(Invocable invocable, List<InvocationArgument> arguments) {
-    if (!isConstructor((NameDeclaration) invocable)) {
-      Type returnType = invocable.getReturnType();
-      if (returnType instanceof IntrinsicReturnType) {
-        List<Type> argumentTypes =
-            arguments.stream()
-                .map(InvocationArgument::getType)
-                .collect(Collectors.toUnmodifiableList());
-        returnType = ((IntrinsicReturnType) returnType).getReturnType(argumentTypes);
+    // Constructors are a special case - they return the type they are invoked on
+    if (isConstructor((NameDeclaration) invocable)) {
+      if (currentType.isClassReference()) {
+        // Calling the constructor on a class reference type returns an instance of that class
+        updateType(((ClassReferenceType) currentType).classType());
       }
-      updateType(returnType);
+
+      return;
     }
+
+    Type returnType = invocable.getReturnType();
+
+    if (returnType instanceof IntrinsicReturnType) {
+      List<Type> argumentTypes =
+          arguments.stream()
+              .map(InvocationArgument::getType)
+              .collect(Collectors.toUnmodifiableList());
+      returnType = ((IntrinsicReturnType) returnType).getReturnType(argumentTypes);
+    }
+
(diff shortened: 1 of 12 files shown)


delphi-frontend/src/main/java/org/sonar/plugins/communitydelphi/api/ast/TypeParameterNode.java
```diff
@@ -30,4 +31,12 @@ public interface TypeParameterNode extends DelphiNode {
...
 
   List<ConstraintNode> getConstraintNodes();
+
+  List<TypeParameter> getTypeParameters();
+
+  interface TypeParameter {
+    NameDeclarationNode getLocation();
+
+    TypeParameterType getType();
+  }
```
Cirras, line 41: This API already exists on `GenericDefinitionNode`.

Cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/NameResolver.java
```diff
@@ -1113,6 +1113,9 @@ private void resolveReturnType(Invocable invocable, List<InvocationArgument> arg
         returnType = ((IntrinsicReturnType) returnType).getReturnType(argumentTypes);
       }
       updateType(returnType);
+    } else if (currentType.isClassReference()) {
```
Cirras, line 1116: Might be easier to read with the conditions flipped around, what do you think?
```java
  private void resolveReturnType(Invocable invocable, List<InvocationArgument> arguments) {
    if (isConstructor((NameDeclaration) invocable)) {
      if (currentType.isClassReference()) {
        // Calling the constructor on a class reference type returns an instance of that class
        updateType(((ClassReferenceType) currentType).classType());
      }
      return;
    }

    Type returnType = invocable.getReturnType();

    if (returnType instanceof IntrinsicReturnType) {
      List<Type> argumentTypes =
          arguments.stream()
              .map(InvocationArgument::getType)
              .collect(Collectors.toUnmodifiableList());
      returnType = ((IntrinsicReturnType) returnType).getReturnType(argumentTypes);
    }

    updateType(returnType);
  }
```

Cirras (changes requested):

delphi-frontend/src/main/java/org/sonar/plugins/communitydelphi/api/ast/TypeParameterNode.java
fourls: Yes - this addition is because `GenericDefinitionNode` combines all the type parameters for all `TypeParameterNode`s into a single list, which stops us from resolving both the type parameters and the name declarations in the correct order (left-to-right, not all name declarations followed by all type parameters). 

I even considered deprecating the version in `GenericDefinitionNode`.

Cirras: Alright, let's compromise for maximum misery - deprecated the old API altogether.

fourls: As discussed, I've kept the old API and changed `TypeParameterNode` to use `GenericDefinitionNode.TypeParameter`.

Cirras (approved):
Looks good. 👍

PR #385, opened 2025-05-13
Fix grammar ambiguity between attributes and interface GUIDs

Cirras:
This PR fixes a significant grammar ambiguity between attributes and interface GUIDs.

Consider the following:
```pas
const
  FooAttribute = class(TCustomAttribute)
  end;

  TBar = interface
    [Foo]
    procedure Baz;
  end;
```

At the moment, the grammar interprets `[Foo]` as a GUID containing constant expression `Foo`.
It should actually be interpreted as a custom attribute on the `Baz` procedure.

In the process of resolving this ambiguity, I discovered some awful details.
```pas
const
  FooAttribute = class(TCustomAttribute)
  end;

  TBar = interface
    [Foo, '{4E89BDD0-B67C-4B4D-86E6-2C0D1DC7B237']
    procedure Baz;
  end;
```

This applies `FooAttribute` to `Baz` and applies the GUID to `TBar`.
Based on this, it's clear that the custom attribute and GUID syntax are really one and the same, and the ambiguity was a result of us pretending they were different nodes in the first place.

There's some gnarly technical details here about who should actually owns the attribute node in a case like this.

Here's how it shook out:
- Any `AttributeListNode` as the first child of an `InterfaceTypeNode` is owned by that parent type node.
- The first declaration (routine/property) within the `InterfaceTypeNode` must look upward to find that attribute list in `RoutineHeadingNode::getAttributeList` and `PropertyNode::getAttributeList`.
- There is no further disambiguation of ownership, even if we can statically determine that the final attribute of the first attribute group of the attribute list is a GUID.

This would have been easier to handle with #261.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **API:** `TypeParameterNode::getTypeParameters` method.
+- **API:** `InterfaceTypeNode::getGuidExpression` method.
+- **API:** `AttributeNode::getExpression` method.
+
+### Deprecated
+
+- **API:** `InterfaceGuidNode` node type.
+- **API:** `InterfaceTypeNode::getGuid` method, use `getGuidExpression` instead.
+- **API:** `DelphiTokenType.GUID`, as the associated `InterfaceGuidNode` is no longer parsed.
 
 ### Fixed
 
 - Name resolution failures on generic routine invocations where later type parameters are constrained by earlier type parameters.
 - Type resolution failures on `as` casts where the type is returned from a routine invocation.
 - Inaccurate type resolution when calling a constructor on a class reference type.
+- Grammar ambiguity causing attributes to be misinterpreted as interface GUIDs.
 
 ## [1.16.0] - 2025-05-09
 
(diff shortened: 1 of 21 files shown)


CHANGELOG.md
fourls, line 1: Changelog needs:

- **API:** `InterfaceTypeNode::getGuidExpression` method.
- **API:** `AttributeNode::getExpression` method.

fourls (changes requested):
Looks good - missing a couple of changelog entries.

Cirras:
> * Any `AttributeListNode` as the first child of an `InterfaceTypeNode` is owned by that parent type node.
> * The first declaration (routine/property) within the `InterfaceTypeNode` must look upward to find that attribute list in `RoutineHeadingNode::getAttributeList` and `PropertyNode::getAttributeList`.
> * There is no further disambiguation of ownership, even if we can statically determine that the final attribute of the first attribute group of the attribute list is a GUID.

I've decided that I really hate what this does to the structure of the AST, so I've made changes to the ownership situation.
Instead of `InterfaceTypeNode` **always** owning the first `AttributeListNode`, it now only owns it if there are no member declarations.

This keeps all of the annoying `AttributeListNode` lookup complexity encapsulated to `InterfaceTypeNodeImpl` (it's either a child or on the first member), and makes the AST structure more consistent.

I also ran across some bugs:
- `RoutineName` rule was raising issues on the entire routine declaration node (including attributes)
- `RoutineNameDeclaration::attributeTypes` were always unknown for implementation-location routines.
- `PropertyNameDeclaration::attributeTypes` were always unknown.

fourls (dismissed):
Looks good! The restructure makes this a LOT better.

fourls:
Happy to merge after a rebase to fix changelog conflicts.

fourls (approved):

PR #387, opened 2025-05-14
Improve test coverage in new AST code

Cirras:
This PR improves test coverage in `AttributeNodeImpl` and `InterfaceTypeNodeImpl`.

### DIFF
--- a/delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/AttributeNodeImplTest.java
+++ b/delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/AttributeNodeImplTest.java
@@ -1,0 +1,124 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2025 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.antlr.ast.node;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+import au.com.integradev.delphi.symbol.occurrence.AttributeNameOccurrenceImpl;
+import au.com.integradev.delphi.symbol.occurrence.NameOccurrenceImpl;
+import au.com.integradev.delphi.utils.files.DelphiFileUtils;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.sonar.plugins.communitydelphi.api.ast.AttributeNode;
+import org.sonar.plugins.communitydelphi.api.symbol.NameOccurrence;
+
+class AttributeNodeImplTest {
+  @Test
+  void testAttribute() {
+    AttributeNode node = parse("[Foo('Bar')]");
+
+    assertThat(node).isNotNull();
+    assertThat(node.isAssembly()).isFalse();
+    assertThat(node.getImage()).isEqualTo("Foo('Bar')");
+    assertThat(node.getNameReference()).isNotNull();
+    assertThat(node.getArgumentList()).isNotNull();
+  }
+
+  @Test
+  void testAttributeNameOccurrence() {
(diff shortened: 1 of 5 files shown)


delphi-frontend/src/test/java/au/com/integradev/delphi/utils/files/DelphiFileUtils.java
```diff
@@ -23,18 +23,34 @@
...
 
 public final class DelphiFileUtils {
   private DelphiFileUtils() {
     // Utility class
   }
 
+  public static DelphiFile parse(String content) {
+    try {
+      Path path = Files.createTempFile(null, ".pas");
+      Files.writeString(path, content, StandardCharsets.UTF_8);
```
fourls, line 46: This should also write a UTF-8 BOM:

```suggestion
     Files.writeString(path, "\uFEFF" + content, StandardCharsets.UTF_8);
```

delphi-frontend/src/test/java/au/com/integradev/delphi/utils/files/DelphiFileUtils.java
```diff
@@ -23,18 +23,34 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collections;
 
 public final class DelphiFileUtils {
   private DelphiFileUtils() {
     // Utility class
   }
 
+  public static DelphiFile parse(String content) {
```
fourls, line 43: It might be helpful to add an overload for `String...`, so we don't always need to do the concatenations and `\n`s ourselves.

delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/AttributeNodeImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+    assertThat(node.getNameReference()).isNotNull();
+    assertThat(node.getTypeNameOccurrence()).isNull();
+    assertThat(node.getConstructorNameOccurrence()).isNull();
+
+    var nameReference = (NameReferenceNodeImpl) node.getNameReference();
+    var constructorNameOccurrence = mock(NameOccurrence.class);
+    var typeNameOccurrence = new AttributeNameOccurrenceImpl(nameReference);
+
+    typeNameOccurrence.setImplicitConstructorNameOccurrence(constructorNameOccurrence);
+    nameReference.setNameOccurrence(typeNameOccurrence);
```
fourls, line 58: I worry that this could get out of sync with the actual way the symbol table executor handles things. Do you think it might be worth adding another utility to run the symbol table visitor over the AST?

fourls (changes requested):
Code change looks good! Definitely a good idea to start testing some of these nodes a little more - there might be some low hanging fruit in terms of coverage that we can pick off over time.

delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/AttributeNodeImplTest.java
Cirras: I'd rather avoid it for now because then we'd have to dump **another** standard library stub into the project, or I guess try to reuse the other one from the symbol table executor tests.

I'm also not totally sure if it's actually better, as we're sort of whitebox-testing the node's behavior here rather than the behavior of name resolution.

fourls: > I'd rather avoid it for now because then we'd have to dump another standard library stub into the project, or I guess try to reuse the other one from the symbol table executor tests.

Fair enough, these tests don't rely too much on the semantic information so it's not a big deal for these anyway.

> I'm also not totally sure if it's actually better, as we're specifically testing the node's behavior here rather than having this added layer of downstream-testing the behavior of name resolution.

Because we don't have any unit tests for NameResolver / NameResolutionHelper, I think it would be good to have downstream-testing. Our API "promise" is actually how these nodes work in the context of the full table, and it would be nice to test that (although I agree they're not pure unit tests).

delphi-frontend/src/test/java/au/com/integradev/delphi/utils/files/DelphiFileUtils.java
Cirras: Done.

delphi-frontend/src/test/java/au/com/integradev/delphi/utils/files/DelphiFileUtils.java
Cirras: `String ...` can't be disambiguated from a single `String`, so we just end up with the one `parse(String ...)` still.
I think that's fine.

delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/AttributeNodeImplTest.java
Cirras: Yeah, that's fair.

It would be ideal to have more pure unit tests for that name resolution code in particular, but that's quite tricky due to the sort of back-and-forth logic between the AST nodes and the name resolution process.
We rely on AST logic during name resolution, but also rely on name resolution for various AST APIs like `getNameOccurrence`, `getNameDeclaration`, `getScope`, `getType`, etc.

If I could wave a magic wand and drag a fully decoupled semantic API out from the current AST APIs, that would be awfully nice. SonarJava has something like that.

fourls (approved):
Looks good!

PR #388, opened 2025-05-23
Handle variant conversions more accurately

cirras:
This one was a real mess.

The initial implementation of variant conversions for invocation resolution was based on the FPC `is_better_candidate_single_variant` routine because their documentation indicated it was reverse-engineered from Delphi's behavior. Unfortunately, it was inaccurate in many ways.

Some interesting problems in the previous implementation:
- Comparisons of integer types were based on type identity, while Delphi bases these comparisons on signedness and value range.
- All real types (as well as `Currency` and `Comp`) were preferred over `Extended`, while Delphi always prefers it over other real types (plus `Currency` and `Comp`, but only when the`Extended` type is larger).
- `Currency` and `Comp` didn't conflict with `Single` or `Extended`, while in Delphi they will always conflict with real types of the same size or smaller.
- Character types (`WideChar`, `AnsiChar`) parameters were considered valid targets for variant conversions, while in Delphi they are not.
- Mixing-and-matching candidates between string, integer, boolean, and real types was allowed, while in Delphi such cases are ambiguous.


Cirras:
Oops. Repeated word in the commit message.
> ... FPC routine
`is_better_candidate_single_variant` routine ...

PR #393, opened 2025-07-02
Migrate from OSSRH to the Central Portal

cirras:
This PR unblocks the July release by migrating from OSSRH (EOL) to the Central Portal.
Repository secrets have already been updated.

Successful run here (including staging/validation, but of course not publishing):
https://github.com/cirras/sonar-delphi/actions/runs/16017755532/job/45187583849


fourls (approved):
All looks good to me, I guess the real test is doing the actual release 😄

PR #397, opened 2025-07-23
Add full support for `HIGHCHARUNICODE`

### LINKED ISSUE #368: Type inferrence for character literals (and HIGHCHARUNICODE handling)
### Prerequisites

### Engine area

Delphi language support

### Improvement description

Related to #116 is the problem of type inference of character literals.

The page on the [HIGHCHARUNICODE](https://docwiki.embarcadero.com/RADStudio/Athens/en/HIGHCHARUNICODE_directive_(Delphi)) directive provides a pretty good explanation of how it's supposed to work
> The {$HIGHCHARUNICODE ON} directive controls the behavior of characters #$80 ... #$FF (#&NoBreak;128 ... #&NoBreak;255).
>
> When HIGHCHARUNICODE is OFF:
> * All decimal #xxx n-digit literals are parsed as AnsiChar.
> * All hexadecimal #$xx 2-digit literals are parsed as AnsiChar.
> * All hexadecimal #$xxxx 4-digit literals are parsed as WideChar.
>
> When HIGHCHARUNICODE is ON:
> * All literals are parsed as WideChar.

This tells us that the type of a character literal depends both on its value *and* the state of the `HIGHCHARUNICODE` directive.

Here's a short example that illustrates how this could make a difference to analysis
```delphi
{$HIGHCHARUNICODE OFF}

procedure Foo(a: AnsiChar); overload;
begin
end;

procedure Foo(a: WideChar); overload;
begin
end;

begin
  Foo(#128); // calls Foo(AnsiChar) (when HIGHCHARUNICODE is OFF)
  Foo(#127); // calls Foo(WideChar)
end.
```

### Rationale

As in #116, this is required for fully accurate type resolution.

fourls:
Fixes #368 by improving `TextLiteralNode::characterEscapeToChar`. Note that the interpretation of "ANSI encoding" is the system native encoding, so it is assumed that the system the Sonar scan runs on has the same native encoding as the system compiling the code.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
@@ -20,6 +20,9 @@ package au.com.integradev.delphi.antlr.ast.node;
 
 import au.com.integradev.delphi.antlr.ast.visitors.DelphiParserVisitor;
 import au.com.integradev.delphi.preprocessor.TextBlockLineEndingMode;
+import au.com.integradev.delphi.utils.CharsetUtils;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.stream.Collectors;
@@ -28,6 +31,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Strings;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.TextLiteralNode;
+import org.sonar.plugins.communitydelphi.api.directive.SwitchDirective.SwitchKind;
 import org.sonar.plugins.communitydelphi.api.token.DelphiTokenType;
 import org.sonar.plugins.communitydelphi.api.type.IntrinsicType;
 import org.sonar.plugins.communitydelphi.api.type.Type;
@@ -51,9 +55,15 @@ public final class TextLiteralNodeImpl extends DelphiNodeImpl implements TextLit
 
   @Override
   public Type getType() {
-    IntrinsicType intrinsic =
-        (getValue().length() == 1) ? IntrinsicType.CHAR : IntrinsicType.STRING;
+    boolean isSingleCharacter = (getValue().length() == 1);
 
+    if (hasAnsiCharacterEscape()) {
+      return isSingleCharacter
+          ? getTypeFactory().getIntrinsic(IntrinsicType.ANSICHAR)
+          : getTypeFactory().getIntrinsic(IntrinsicType.ANSISTRING);
+    }
+
+    IntrinsicType intrinsic = isSingleCharacter ? IntrinsicType.CHAR : IntrinsicType.STRING;
     return getTypeFactory().getIntrinsic(intrinsic);
   }
 
@@ -167,7 +177,42 @@ public final class TextLiteralNodeImpl extends DelphiNodeImpl implements TextLit
     return imageBuilder.toString();
   }
 
-  private static char characterEscapeToChar(String image) {
+  private boolean isHighCharUnicode() {
+    return getAst()
+        .getDelphiFile()
+        .getCompilerSwitchRegistry()
+        .isActiveSwitch(SwitchKind.HIGHCHARUNICODE, getTokenIndex());
+  }
+
+  private boolean hasAnsiCharacterEscape() {
+    if (isHighCharUnicode()) {
+      return false;
+    }
+
+    return getChildren().stream()
+        .filter(child -> child.getTokenType() == DelphiTokenType.CHARACTER_ESCAPE_CODE)
+        .map(DelphiNode::getImage)
+        .mapToInt(TextLiteralNodeImpl::parseCharacterEscape)
+        .anyMatch(TextLiteralNodeImpl::isAnsiCodePoint);
+  }
+
+  private char characterEscapeToChar(String image) {
+    int codePoint = parseCharacterEscape(image);
+
+    if (!isHighCharUnicode() && isAnsiCodePoint(codePoint)) {
+      var buffer = ByteBuffer.allocate(1).put((byte) codePoint).flip();
+      return getAnsiCharset().decode(buffer).get();
+    }
+
+    return (char) codePoint;
+  }
+
+  private Charset getAnsiCharset() {
+    Charset ansiCharset = getAst().getDelphiFile().getAnsiCharset();
+    return ansiCharset != null ? ansiCharset : CharsetUtils.nativeCharset();
+  }
+
+  private static int parseCharacterEscape(String image) {
     image = image.substring(1);
     int radix = 10;
 
@@ -181,12 +226,15 @@ public final class TextLiteralNodeImpl extends DelphiNodeImpl implements TextLit
         image = image.substring(1);
         break;
       default:
-        // do nothing
+        // Do nothing
     }
 
     image = StringUtils.remove(image, '_');
+    return Integer.parseInt(image, radix);
+  }
 
-    return (char) Integer.parseInt(image, radix);
+  private static boolean isAnsiCodePoint(int codePoint) {
+    return codePoint >= 128 && codePoint <= 255;
   }
 
   @Override
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectFactory.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectFactory.java
@@ -58,6 +58,7 @@ final class DelphiProjectFactory {
     project.setLibraryPath(createLibraryPathDirectories(state, projectDirectory));
     project.setBrowsingPath(createBrowsingPathDirectories(state, projectDirectory));
     project.setUnitAliases(createUnitAliases(state));
+    project.setCodePage(createCodePage(state));
     project.setSourceFiles(sourceFiles);
 
     return project;
@@ -149,6 +150,20 @@ final class DelphiProjectFactory {
     return Collections.unmodifiableMap(result);
   }
 
+  private static Integer createCodePage(MSBuildState state) {
+    String codePage = state.getProperty("DCC_CodePage");
+    if (codePage == null) {
+      return null;
+    }
+
+    try {
+      return Integer.parseInt(codePage);
+    } catch (NumberFormatException e) {
+      LOG.warn("Invalid DCC_CodePage value: {}", codePage);
+      return null;
+    }
+  }
+
   private static List<String> propertyList(String value) {
     if (value == null) {
       return Collections.emptyList();
@@ -165,6 +180,7 @@ final class DelphiProjectFactory {
     private List<Path> libraryPathDirectories = Collections.emptyList();
     private List<Path> browsingPathDirectories = Collections.emptyList();
     private Map<String, String> unitAliases = Collections.emptyMap();
+    private Integer codePage;
 
     private void setDefinitions(Set<String> definitions) {
       this.definitions = ImmutableSortedSet.copyOf(String.CASE_INSENSITIVE_ORDER, definitions);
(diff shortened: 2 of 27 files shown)


fourls:
One thing I forgot to note - I removed support for binary character escapes (e.g. `#%11101110`) since as far as I can tell Delphi does not recognise them.

zaneduffield:
The system encoding isn't what the compiler is using. The compiler is using the configured codepage, which *may* be 0, which means the system encoding.

zaneduffield:
Actually... let me check that.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,26 +170,38 @@ private String createSingleLineValue() {
...
-      case '%':
-        radix = 2;
-        image = image.substring(1);
-        break;
-      default:
-        // do nothing
+    if (image.charAt(0) == '$') {
+      radix = 16;
+      image = image.substring(1);
     }
```
cirras, line 191: Let's keep the handling for binary character escapes even though they're currently not syntactically valid.
Our grammar allows it, and it's arguably a bug that the compiler doesn't currently recognize them. I think it's likely a future Delphi version will fix that.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,26 +170,38 @@ private String createSingleLineValue() {
...
     image = StringUtils.remove(image, '_');
+    char character = (char) Integer.parseInt(image, radix);
 
-    return (char) Integer.parseInt(image, radix);
+    if (isHighCharUnicode() || character > 255) {
+      // With HIGHCHARUNICODE ON, all escapes are interpreted as UTF-16.
+      // Escapes above 255 are always interpreted as UTF-16.
+      return character;
+    } else {
+      // With HIGHCHARUNICODE OFF, escapes between 0-255 are interpreted in the system code page.
```
cirras, line 201: According to the `HIGHCHARUNICODE` documentation, only the `#128`-`#255` range is affected.

This wouldn't seem to matter, if you're only thinking about single-byte ANSI codepages that are supersets of ASCII.
However, there's multi-byte ANSI codepages that *aren't* supersets of ASCII (I think?), and it seems like this aspect of the behavior would matter for interpreting those.

cirras: With that being said, I've looked at the Shift_JIS character table and found that the first 127 characters are still ASCII.
Maybe it really doesn't matter and I'm just getting muddled up with the fact that there are codepages that aren't **binary** supersets of ASCII.

Even so, we should probably follow what the documentation says.

delphi-frontend/src/test/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImplTest.java
```diff
@@ -59,22 +66,45 @@ void testMultilineImage() {
     assertThat(node.isMultiline()).isTrue();
   }
 
-  @Test
-  void testGetImageWithCharacterEscapes() {
-    TextLiteralNodeImpl node = new TextLiteralNodeImpl(DelphiLexer.TkTextLiteral);
+  @ParameterizedTest(name = "HIGHCHARUNICODE = {0}")
+  @ValueSource(booleans = {true, false})
+  void testGetImageWithCharacterEscapes(boolean highCharUnicode) {
```
cirras, line 71: The switch also affects the type of the expression - not just the way the string is interpreted.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,26 +170,38 @@ private String createSingleLineValue() {
...
 
-  private static char characterEscapeToChar(String image) {
+  private boolean isHighCharUnicode() {
+    return getAst()
+        .getDelphiFile()
+        .getCompilerSwitchRegistry()
+        .isActiveSwitch(SwitchKind.HIGHCHARUNICODE, getTokenIndex());
+  }
+
+  public Charset getAnsiCharset() {
```
cirras, line 180: I think we should probably have a `CharsetUtils` or something in frontend.
I'd also call these `nativeCharset`.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,26 +170,38 @@ private String createSingleLineValue() {
...
+    return Charset.forName(System.getProperty("native.encoding"));
```
cirras, line 181: Assuming that the compiler is actually using the configured codepage to interpret these escapes, we should:
- read `DCC_Codepage` from dproj files
- emit a warning if conflicting `DCC_Codepage` values are found
- expose an analyzer property to override `DCC_Codepage` and ignore it altogether
- fall back to the system's native encoding if none of these are provided

cirras (changes requested):

cirras (changes requested):

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,26 +170,38 @@ private String createSingleLineValue() {
...
 
-    return (char) Integer.parseInt(image, radix);
+    if (isHighCharUnicode() || character > 255) {
+      // With HIGHCHARUNICODE ON, all escapes are interpreted as UTF-16.
+      // Escapes above 255 are always interpreted as UTF-16.
+      return character;
+    } else {
+      // With HIGHCHARUNICODE OFF, escapes between 0-255 are interpreted in the system code page.
+      var buffer = ByteBuffer.allocate(1).put((byte) character).flip();
+      return getAnsiCharset().decode(buffer).get();
```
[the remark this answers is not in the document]
fourls: As discussed, the compiler does always interpret the escape as the byte representation, not as a codepoint. If the byte representation doesn't end up as a valid character, it uses the replacement char for the codepage, e.g. `U+FFFD` for UTF-8 or `?` for most single-byte code pages.

cirras:
@fourls, as discussed I have actioned the feedback points from this review.
Please have a look over my changes.

delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectHelper.java
```diff
@@ -183,6 +188,61 @@ private Set<String> getPredefinedConditionalDefines() {
...
+  private Charset resolveAnsiCharset() {
+    Integer configuredCodePage = getCodePageFromSettings();
+    if (configuredCodePage != null) {
+      return charsetForCodePage(configuredCodePage);
+    }
+
+    List<Integer> projectCodePages =
+        projects.stream()
+            .map(DelphiProject::getCodePage)
+            .filter(Objects::nonNull)
```
fourls: The absence of a `DCC_CodePage` override doesn't necessarily mean the value isn't relevant. I think we should be treating null values as `0` (default) instead of filtering them out, e.g. if I have one project with code page `65001` and one project with no configured code page, then it shouldn't act differently to if project 2 was code page `0`. 

We could do this either in `DelphiProject::getCodePage` or here via another `map()`.

delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectFactory.java
```diff
@@ -149,6 +150,20 @@ private static Map<String, String> createUnitAliases(MSBuildState state) {
...
+  private static Integer createCodePage(MSBuildState state) {
+    String codePage = state.getProperty("DCC_Codepage");
+    if (codePage == null) {
+      return null;
+    }
+
+    try {
+      return Integer.parseInt(codePage);
+    } catch (NumberFormatException e) {
+      LOG.warn("Invalid DCC_Codepage value: {}", codePage);
```
fourls: Tiny nitpick: the canonical casing is `DCC_CodePage`.

delphi-frontend/src/main/java/au/com/integradev/delphi/utils/CharsetUtils.java
```diff
@@ -0,0 +1,203 @@
...
+package au.com.integradev.delphi.utils;
+
+import com.google.common.collect.ImmutableMap;
+import java.nio.charset.Charset;
+import java.util.Map;
+import org.sonar.plugins.communitydelphi.api.type.CodePages;
+
+public final class CharsetUtils {
+  private static final String NATIVE_ENCODING_PROPERTY = "native.encoding";
+  private static final Map<Integer, String> CODE_PAGE_CHARSET_NAMES =
```
fourls, line 28: Should we use constants from `CodePages` here?

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/node/TextLiteralNodeImpl.java
```diff
@@ -167,7 +177,43 @@ private String createSingleLineValue() {
...
+    return getChildren().stream()
+        .filter(child -> child.getTokenType() == DelphiTokenType.CHARACTER_ESCAPE_CODE)
+        .map(DelphiNode::getImage)
+        .mapToInt(TextLiteralNodeImpl::parseCharacterEscape)
+        .anyMatch(TextLiteralNodeImpl::isAnsiCodePoint);
+  }
+
+  private char characterEscapeToChar(String image) {
+    int codePoint = parseCharacterEscape(image);
+    char character = (char) codePoint;
```
fourls: This variable isn't really necessary - just cast `codePoint` to `char` when returning it.

delphi-frontend/src/main/java/au/com/integradev/delphi/utils/CharsetUtils.java
cirras: I'd prefer not to use the constants in this hellish table of codepage -> charset names.

I kind of regret making that `CodePages` class public in the API and would like to make it private in 2.0.0. `CP_1252` in particular has no excuse for existing.

In fact, I may raise a separate PR to introduce a new private copy of `CodePages` and deprecate the public existing one for removal.

delphi-frontend/src/main/java/au/com/integradev/delphi/msbuild/DelphiProjectFactory.java
cirras: Oops! Good catch.

cirras:
@fourls, addressed your feedback points. Can you have a look?

delphi-frontend/src/test/java/au/com/integradev/delphi/msbuild/DelphiProjectHelperTest.java
```diff
@@ -317,6 +318,121 @@ void testInvalidCompilerVersion() {
...
+    DelphiProjectHelper delphiProjectHelper =
+        new DelphiProjectHelper(settings, fs, environmentVariableProvider);
+
+    assertThat(delphiProjectHelper.getAnsiCharset()).isEqualTo(StandardCharsets.ISO_8859_1);
+  }
+
+  @Test
+  void testNoConfiguredCodePageFallsBackToProjectCodePage() {
+    addInputFile(PROJECTS_PATH + "CodePageProject/Utf8.dproj");
+    when(settings.get(DelphiProperties.CODE_PAGE_KEY)).thenReturn(Optional.of("  \t  "));
```
fourls: Is this test correct and necessary? While the implementation looks like it tests the same thing as `testInvalidConfiguredCodePageFallsBackToProjectCodePage`, the name implies that there should be no code page setting configured?

cirras: Oops, I remember the quadruple copy-paste that yielded this test... It definitely doesn't need to exist, we already cover this path in `testProjectCodePageProvidesAnsiCharset`.

fourls (commented):
✅ Looks good!

---

_fake approval brought to you by "guy who left his PR unfinished for a year"_

cirras (approved):
Looks good to me!

PR #399, opened 2025-07-24
Implement `ObjectPassedAsInterface` check

### LINKED ISSUE #99: New rule: Class instance variables should not be passed as an interface parameter
### Prerequisites

### Suggested rule title

Class instance variables should not be passed as an interface parameter

### Rule description

This rule picks up cases where a variable containing an object instance (e.g. `TObject`) is passed to a routine that expects an interface (e.g. `IInterface`). The interface reference causes the object to be reference counted and freed at the end of the routine, meaning that the variable is pointing to freed memory when control is returned to the calling code.

```delphi
type
  IFooIntf = interface
  end;

  TFooImpl = class(TInterfacedObject, IFooIntf)
  end;

procedure DoThing(Foo: IFooIntf);
begin
  // ...
end;

procedure Run;
var
  FooImpl: TFooImpl;
begin
  FooImpl := TFooImpl.Create;
  DoThing(FooImpl); // Noncompliant - performs reference counting and destroys object before returning control
  FooImpl.Free; // This will cause a double free
end;
```

### Rationale

This is a common "gotcha" as it is a frequent pattern in other languages. It is also a statically verifiable case in which access violations are likely.

fourls:
Fixes #99.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `ObjectPassedAsInterface` analysis rule, which detects object references that are passed directly as
+  an interface to a routine.
+
 ### Changed
 
 - `EmptyBlock` now ignores all empty blocks containing an explanatory comment.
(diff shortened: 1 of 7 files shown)


cirras (changes requested):
Looks good, just needs a CHANGELOG entry.

PR #400, opened 2025-07-25
Implement `TooManyDefaultParameters` check

### LINKED ISSUE #74: New rule: Methods should not have too many default parameters
### Prerequisites

### Suggested rule title

Methods should not have too many default parameters

### Rule description

This rule would identify methods with more than a configured maximum number of default parameters. This would be a user-configurable property.

### Rationale

Default parameters can be useful in cases when there is a sensible default value. They can, however, damage API interfaces when used to make a method's behaviour overly configurable - a method with too many default parameters is confusing, unpredictable at a glance, unintuitive to use, and likely overloaded in functionality. 

Using multiple methods, method overloads, or configuration objects would be a more maintainable approach and would keep the API cleaner.

fourls:
Fixes #74.

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/TooManyDefaultParametersCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/TooManyDefaultParametersCheck.java
@@ -1,0 +1,75 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2025 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+import org.sonar.plugins.communitydelphi.api.ast.FormalParameterNode.FormalParameterData;
+import org.sonar.plugins.communitydelphi.api.ast.RoutineDeclarationNode;
+import org.sonar.plugins.communitydelphi.api.ast.RoutineImplementationNode;
+import org.sonar.plugins.communitydelphi.api.ast.RoutineNameNode;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheck;
+import org.sonar.plugins.communitydelphi.api.check.DelphiCheckContext;
+import org.sonar.plugins.communitydelphi.api.type.Parameter;
+
+@Rule(key = "TooManyDefaultParameters")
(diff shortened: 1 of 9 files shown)


delphi-checks/src/main/java/au/com/integradev/delphi/checks/TooManyDefaultParametersCheck.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final int DEFAULT_MAXIMUM = 2;
+
+  @RuleProperty(
+      key = "max",
+      description = "Maximum authorized number of default parameters",
+      defaultValue = "" + DEFAULT_MAXIMUM)
+  public int max = DEFAULT_MAXIMUM;
+
+  @Override
+  public DelphiCheckContext visit(RoutineImplementationNode routine, DelphiCheckContext context) {
```
cirras, line 40: What about interface or abstract method declarations?

I feel like we should really be looking directly at the declaration nodes and only raising issues on the implementation node when it's an implementation-local routine.

delphi-checks/src/test/java/au/com/integradev/delphi/checks/TooManyDefaultParametersCheckTest.java
```diff
@@ -0,0 +1,219 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2019 Integrated Application Development
```
cirras, line 3:
```suggestion
 * Copyright (C) 2025 Integrated Application Development
```

delphi-checks/src/main/java/au/com/integradev/delphi/checks/TooManyDefaultParametersCheck.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2019 Integrated Application Development
```
cirras, line 3:
```suggestion
 * Copyright (C) 2025 Integrated Application Development
```

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/TooManyDefaultParameters.html
```diff
@@ -0,0 +1,20 @@
+<h2>Why is this an issue?</h2>
+<p>
+  Default parameters can be useful in cases when there is a sensible default value. They can,
+  however, damage API interfaces when used to make a method's behaviour overly configurable
+  a method with too many default parameters is confusing, unpredictable at a glance, unintuitive
+  to use, and likely overloaded in functionality.
```
cirras, line 6: That second sentence looks like a typo'd run-on.

delphi-checks/src/main/resources/org/sonar/l10n/delphi/rules/community-delphi/TooManyDefaultParameters.html
```diff
@@ -0,0 +1,20 @@
...
+  appropriate solution.
+</p>
+<h2>How to fix it</h2>
+<p>
+  Refactor this routine by re-evaluating the need for the default values, and either:
+</p>
+<ul>
+  <li>Removing the unnecessary default values,</li>
+  <li>Adding multiple overloads with different groups of required parameters, or</li>
+  <li>Using a configuration object approach instead</li>
```
cirras, line 19: Do we prefer the term "configuration object" for this, or should we say something like "Grouping parameters into an object"?

cirras (changes requested):

cirras (changes requested):
Looks good, we just need CHANGELOG entries for the new rule and the changes to `TooManyParameters`.

cirras (dismissed):
Looks good! Just needs a rebase.

cirras (approved):

PR #402, opened 2025-08-06
Improve type comparisons between classes and interfaces

cirras:
This PR fixes a type comparison inaccuracy where class-to-class upcasts weren't always preferred over class-to-interface upcasts.

The downstream effects were incorrect overload resolution, and false positives in rules like `ObjectPassedAsInterface` where the rule would think an object was being passed to an interface parameter on the incorrectly-selected overload.


fourls (approved):
Looks good!

PR #406, opened 2025-09-07
Fail the release workflow if the plugin jar is not found

cirras:
Now that immutable releases are available on GitHub, I'm enabling them for this repository.

I've made this tweak to the release workflow to ensure the release workflow can't generates a release without the plugin jar attached. Such a release wouldn't be manually fixable, now that the releases and their artifacts are immutable.


fourls (approved):
Looks good!

PR #433, opened 2026-04-09
Add support for 'noreturn' keyword introduced in D13.

### LINKED ISSUE #410: Support `noreturn` routine directive
### Prerequisites

### Engine area

Delphi language support

### Improvement description

Delphi 13 has introduced a new `noreturn` routine directive.

It indicates to the compiler that "the call is going to raise an exception and not complete regular execution."

Applying the `noreturn` directive tells the compiler that it can make some optimizations and skip some warnings at the callsite (like [W1035](https://docwiki.embarcadero.com/RADStudio/Florence/en/W1035_Return_value_of_function_%27%25s%27_might_be_undefined_(Delphi))) since regular execution will not continue after the call.

A good example of a `noreturn` routine is the `OutOfMemoryError` procedure in the standard library. It unconditionally raises a cached `OutOfMemoryException`.

We'll need to:
- parse the new syntax (`noreturn` token)
- extend the CFG to treat calls to `noreturn` routines like unconditional exception raises
- **API:** extend the `RoutineDirective` enum

## See
- [The `noreturn` Directive](https://docwiki.embarcadero.com/RADStudio/Florence/en/Declarations_and_Statements_(Delphi)#The_noreturn_Directive)

### Rationale

Delphi 13 support.

Bazooper-blip:


cirras (dismissed):
Made a few changes:
- Changed the rule name to `NoreturnContract`.
- Improved modelling of exception raises in the CFG to better support the rule you added.
- Fixed the rule implementation to walk the CFG correctly. (tests are now passing)

I'll merge shortly - thanks for your contribution. 👍

cirras (approved):

PR #436, opened 2026-04-15
Add support for implicit Self in Initialize and Finalize operators

### LINKED ISSUE #416: Support implicit `Self` parameter in `Initialize` and `Finalize`
### Prerequisites

### Engine area

Delphi language support

### Improvement description

Delphi 13 now implicitly declares `Self` in `Initialize` and `Finalize` operators, so the user doesn't need to manually declare it as a parameter.

This should only require a minor tweak in the `SymbolTableVisitor`.

## See
- [Implicit Self parameter declaration](https://docwiki.embarcadero.com/RADStudio/Florence/en/Custom_Managed_Records#Implicit_Self_parameter_declaration)

### Rationale

Delphi 13 support.

Bazooper-blip:
Delphi 13 automatically declares Self in Initialize and Finalize operators for custom managed records. Previously developers had to manually declare it as a parameter.

Updates findSelfType() in SymbolTableVisitor to skip classOf() wrapping for these specific record operators, so Self resolves to the record type directly.

Closes #416

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for `noreturn` routines, introduced in Delphi 13.
 - Support for `NameOf` intrinsic, introduced in Delphi 13.
+- Support for implicit `Self` in `Initialize` and `Finalize` operators, introduced in Delphi 13.
 - `NoreturnContract` analysis rule, which flags `noreturn` routines that return normally.
 
 ## [1.18.3] - 2025-11-11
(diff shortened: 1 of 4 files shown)


cirras:
I've made some changes:
- Tidied the implementation of `SymbolTableVisitor::findSelfType`. 
  - In the process, fixed a subtle bug where all class operator implementations erroneously had access to an implicit `Self` in the same way class methods do.
- Rewrote `testRecordInitializeFinalizeSelfTypes`:
  - Test case 1: "parameter named `Self` provided".
  - Test case 2: "no parameter provided" 
  - Removed inappropriate "regression test" which was irrelevant to testing `Initialize`/`Finalize` operator behavior.
- Removed overly-specific description of changes from your commit message body.

cirras (approved):

PR #441, opened 2026-05-12
Improve `NoreturnContractCheck`

jgardn3r:
- **Introduce `RoutineExit` type to control flow graph types**
- **Remove `ControlFlowGraph::getExitBlock`**
- **Fix CFG construction of `Continue` and `Break` around `try finally`**
- **Ensure `RoutineExit` block is always id 0**
- **Replace AssertJ `withFailMessage` with `as` in CFG `BlockChecker`**
- **Add `ExceptionalRoutineExit` to appropriate `ControlFlowGraph`s**
- **Improve `NoreturnContractCheck`**

### DIFF
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
@@ -20,6 +20,7 @@ package au.com.integradev.delphi.checks;
 
 import au.com.integradev.delphi.cfg.api.Block;
 import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
+import au.com.integradev.delphi.cfg.api.ExceptionalRoutineExit;
 import au.com.integradev.delphi.cfg.api.Finally;
 import au.com.integradev.delphi.cfg.api.PossibleException;
 import au.com.integradev.delphi.cfg.api.RoutineExit;
@@ -87,6 +88,6 @@ public class NoreturnContractCheck extends DelphiCheck {
   }
 
   private static boolean notFinallyOrExit(Block block) {
-    return !(block instanceof Finally || block instanceof RoutineExit);
+    return !(block instanceof Finally || block instanceof ExceptionalRoutineExit);
   }
 }
--- a/delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
+++ b/delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
@@ -22,6 +22,7 @@ import au.com.integradev.delphi.cfg.api.Block;
 import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
 import au.com.integradev.delphi.cfg.api.Finally;
 import au.com.integradev.delphi.cfg.api.PossibleException;
+import au.com.integradev.delphi.cfg.api.RoutineExit;
 import au.com.integradev.delphi.cfg.api.Terminated;
 import au.com.integradev.delphi.cfg.api.UnknownException;
 import au.com.integradev.delphi.cfg.block.TerminatorKind;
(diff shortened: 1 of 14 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/block/ProtoBlockFactory.java
```diff
@@ -18,16 +18,7 @@
...
-import au.com.integradev.delphi.cfg.api.Branch;
-import au.com.integradev.delphi.cfg.api.Cases;
-import au.com.integradev.delphi.cfg.api.Finally;
-import au.com.integradev.delphi.cfg.api.Halt;
-import au.com.integradev.delphi.cfg.api.Linear;
-import au.com.integradev.delphi.cfg.api.PossibleException;
-import au.com.integradev.delphi.cfg.api.RoutineExit;
-import au.com.integradev.delphi.cfg.api.UnconditionalJump;
-import au.com.integradev.delphi.cfg.api.UnknownException;
+import au.com.integradev.delphi.cfg.api.*;
```
cirras: Let's use explicit imports (you may need an IDE configuration change.)

delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
```diff
@@ -18,18 +18,10 @@
...
 package au.com.integradev.delphi.checks;
 
-import au.com.integradev.delphi.cfg.api.Block;
-import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
-import au.com.integradev.delphi.cfg.api.Finally;
-import au.com.integradev.delphi.cfg.api.PossibleException;
-import au.com.integradev.delphi.cfg.api.RoutineExit;
-import au.com.integradev.delphi.cfg.api.Terminated;
-import au.com.integradev.delphi.cfg.api.UnknownException;
+import au.com.integradev.delphi.cfg.api.*;
```
cirras: Let's use explicit imports.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
```diff
@@ -18,18 +18,10 @@
...
 import au.com.integradev.delphi.cfg.block.TerminatorKind;
 import au.com.integradev.delphi.utils.ControlFlowGraphUtils;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.*;
```
cirras: Let's use explicit imports.

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphImpl.java
```diff
@@ -68,6 +66,20 @@ public void prune() {
 
       blocks.forEach(inactiveBlocks::remove);
     } while (!inactiveBlocks.isEmpty());
+
+    // Remove the exceptional exit if there are no paths to it
+    Block exceptionalExitBlock =
+        ListUtils.reverse(blocks).stream()
```
cirras: Can we use the guava `Lists.reverse` instead? I don't like the idea of using these super specific collection utils from sonarsource analyzer commons (and trust the long-term stability of Guava's API much more.)

delphi-frontend/src/main/java/au/com/integradev/delphi/cfg/ControlFlowGraphImpl.java
```diff
@@ -18,17 +18,15 @@
  */
 package au.com.integradev.delphi.cfg;
 
-import au.com.integradev.delphi.cfg.api.Block;
-import au.com.integradev.delphi.cfg.api.ControlFlowGraph;
-import au.com.integradev.delphi.cfg.api.Terminated;
-import au.com.integradev.delphi.cfg.api.UnconditionalJump;
+import au.com.integradev.delphi.cfg.api.*;
```
cirras: Let's use explicit imports.

cirras (changes requested):

delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
cirras, line 1: The previous implementation passes all tests if we simply change the implementation of `notFinallyOrExit`:
```java
private static boolean notFinallyOrExit(Block block) {
  return !(block instanceof Finally || block instanceof ExceptionalRoutineExit);
}
```

What's the benefit of the full-on implementation rejig?
If there's some case that this handles that the previous implementation fundamentally doesn't, can we add unit tests to demonstrate that?

cirras (changes requested):

delphi-checks/src/main/java/au/com/integradev/delphi/checks/NoreturnContractCheck.java
jgardn3r: The rejig made more sense to me. After playing with the existing implementation a bit more, I think the change you suggested is sufficient. I will revert it to that.

cirras (approved):
Looks good! 👍

PR #442, opened 2026-05-22
Report metrics/duplications/coverage correctly on `MAIN` vs `TEST` input files

cirras:
This PR corrects some little problems with how we handle `MAIN` vs `TEST` sources.
- It looks like the "lines" metric is forced off on `TEST` sources, and the official analyzers don't report metrics for test sources.
- The sonar scanner emits warnings when duplications are reported on `TEST` sources.
- Coverage should only be reported on `MAIN` sources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


fourls (changes requested):
Looks good, just needs a changelog entry.

fourls (approved):
Looks good!

PR #444, opened 2026-05-26
Fix `RedundantInherited` quickfix removing too much surrounding code

### LINKED ISSUE #430: Quickfix for Redundant Inherited removes too much when in else clause
### SonarDelphi version

1.18.3

### SonarQube version

_No response_

### Issue description

The quickfix for this "Redundant inherited statement should be removed"
```
procedure TForm47.Button1Click(Sender: TObject);
begin
  if Sender = Button1 then
  begin
    // do something
  end
  else
    inherited;
end;
```

is

```
procedure TForm47.Button1Click(Sender: TObject);
begin

end;
```

Which seems a tad excessive.

### Steps to reproduce

Just create a button event as in the description above.

_No response_

cirras:
This PR fixes some overly-eager quickfix deletions in `RedundantInherited`.

Fixes #430.

### DIFF
--- a/delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantInheritedCheckTest.java
+++ b/delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantInheritedCheckTest.java
@@ -211,6 +211,135 @@ class RedundantInheritedCheckTest {
   }
 
   @Test
+  void testThenBranchShouldAddQuickFix() {
+    CheckVerifier.newVerifier()
+        .withCheck(new RedundantInheritedCheck())
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendDecl("type")
+                .appendDecl("  TBase = class(TObject);")
+                .appendDecl("  TChild = class(TBase)")
+                .appendDecl("    procedure MyProcedure(Sender: TObject);")
+                .appendDecl("  end;")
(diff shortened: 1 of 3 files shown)


delphi-checks/src/test/java/au/com/integradev/delphi/checks/RedundantInheritedCheckTest.java
```diff
@@ -210,6 +248,130 @@ void testNotOverridingMethodShouldAddIssues() {
...
+        .onFile(
+            new DelphiTestUnitBuilder()
+                .appendDecl("type")
+                .appendDecl("  TBase = class(TObject);")
+                .appendDecl("  TChild = class(TBase)")
+                .appendDecl("    procedure MyProcedure(Sender: TObject);")
+                .appendDecl("  end;")
+                .appendImpl("procedure TChild.MyProcedure(Sender: TObject);")
+                .appendImpl("begin")
+                .appendImpl("  while Sender = nil do")
```
jgardn3r, line 368: There are other statements that can have a "direct" statement. E.g., `case` and `on`. Should we add tests for those too?

cirras: Sure.

jgardn3r (approved):
Looks good!

PR #440, opened 2026-04-22
Add support for `if` expressions

Bazooper-blip:
Resolves #407. Introduces parsing, type resolution, and control-flow modelling of Delphi 13's conditional `if` operator (e.g. `X := if Foo then Bar else Baz`). The LUB of the two branch types is used as the expression's type, and the branches are modelled as mutually-exclusive paths in the CFG.

PR #446, opened 2026-05-27
Add support for `if` expressions

### LINKED ISSUE #407: Support `if` expressions
### Prerequisites

### Engine area

Delphi language support

### Improvement description

We'll need to:
- parse the new syntax

## Example
```pas
```

## See

### Rationale

Delphi 13 support.

Bazooper-blip:
Resolves #407. Introduces parsing, type resolution, and control-flow modelling of Delphi 13's conditional `if` operator (e.g. `X := if Foo then Bar else Baz`). The branches are modelled as mutually-exclusive paths in the CFG.

The expression's type is the LUB of the then/else branch types: exact matches and unknown-side fallbacks are returned directly; if only one branch's type is assignable to the other, the more permissive one wins; if both are bidirectionally assignable, the wider one (higher conversion score) is chosen; for unrelated struct types, a shared ancestor in the parent chain is used; otherwise the result is the unknown type.

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ExpressionTypeResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ExpressionTypeResolver.java
@@ -36,6 +36,7 @@ import org.sonar.plugins.communitydelphi.api.ast.BinaryExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.CommonDelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
+import org.sonar.plugins.communitydelphi.api.ast.IfExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.NameReferenceNode;
 import org.sonar.plugins.communitydelphi.api.ast.Node;
 import org.sonar.plugins.communitydelphi.api.ast.PrimaryExpressionNode;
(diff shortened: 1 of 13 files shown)


cirras (commented):
I don't understand why you closed #440 and opened a new PR. If it's because there was a merge commit in that PR, that could have easily been cleaned up by rewriting the PR branch. 🤔

cirras (changes requested):
> Resolves https://github.com/integrated-application-development/sonar-delphi/issues/407. Introduces parsing, type resolution, and control-flow
modelling of Delphi 13's conditional `if` operator
(e.g. `X := if Foo then Bar else Baz`). The branches are modelled as
mutually-exclusive paths in the CFG.
> 
> The expression's type is the LUB of the then/else branch types:
exact matches and unknown-side fallbacks are returned directly; if
only one branch's type is assignable to the other, the more
permissive one wins; if both are bidirectionally assignable, the
wider one (higher conversion score) is chosen; for unrelated struct
types, a shared ancestor in the parent chain is used; otherwise
the result is the unknown type.

This whole commit message body looks like an AI-generated summary of implementation minutia and is not useful.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ExpressionTypeResolver.java
```diff
@@ -93,6 +94,58 @@ public Type resolve(UnaryExpressionNode expression) {
     }
   }
 
+  public Type resolve(IfExpressionNode expression) {
```
cirras, line 98: For the third time, this implementation is simply incorrect. I won't be able to accept this PR without accurate type resolution.

### Please read the [Embarcadero documentation](https://docwiki.embarcadero.com/RADStudio/Florence/en/Conditional_Operators_(Delphi)).

# **... and test your assumptions against the Delphi 13 compiler.**

cirras (changes requested):

joachimmarder:
@cirras : Any chance that you review this pull request soon?

cirras:
Closed in favor of #450.

PR #448, opened 2026-06-09
Handle empty `DCC_CodePage` values without emitting a warning

cirras:
This PR fixes a bug in #397 where we mishandled `DCC_CodePage` values.

It's easy to forget, but we handle empty and null values equivalently when handling MSBuild properties, as that's how MSBuild itself works. It's impossible for `MSBuildState#getProperty` to return a null value.

This was causing erroneous warnings to be emitted for projects that defined an empty `DCC_CodePage` property, or didn't define one at all.


fourls (approved):
Looks good!

PR #449, opened 2026-07-24
Add support for inline if expressions

### LINKED ISSUE #407: Support `if` expressions

Daniel-Tr:
Resolves #407.

Delphi 13 adds an `if <cond> then <expr> else <expr>` conditional
expression. This adds parsing for it, models it as a proper branch in
the CFG (only one side actually runs at runtime, so it isn't
evaluated the way a range or array-constructor expression is), and
resolves its type as the compiler's least-upper-bound of the two
branches.

The LUB isn't a "pick whichever branch converts better" comparison -
that can only ever return one of the two input types, but Delphi's
actual promotion rules sometimes land on neither (e.g. `Byte` and
`Word` resolve to `Integer`, not `Word`), and unrelated class
branches need a common-ancestor walk rather than a closest-match
pick. `ExpressionTypeResolverTest` encodes the promotion rules as a
table, with every expected result checked against `dcc64` directly
(assigning the expression to an incompatible variable and reading the
type off the resulting E2010 error - see the class comment) rather
than inferred from the existing overload-resolution machinery.

Also counts as a decision point in cyclomatic/cognitive complexity,
same as an `if` statement.

This change has mostly been created by Claude and reviewed manually.
We compiled, installed, and beta-tested this change in our CI environment 
against productive code (that fails with the current published version).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for `if` conditional expressions, introduced in Delphi 13.
+
 ## [1.19.0] - 2026-06-09
 
 ### Added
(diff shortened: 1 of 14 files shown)


cirras:
Closed in favor of #450.

PR #450, opened 2026-08-03
Add support for `if` expressions

### LINKED ISSUE #407: Support `if` expressions

cirras:
This PR implements `if` expressions.

## Embarcadero Documentation

The language feature is documented by Embarcadero [here](https://docwiki.embarcadero.com/RADStudio/Florence/en/Conditional_Operators_(Delphi)).

They even kindly provided a table of operand type combinations with the corresponding result type.

| Type 1 | Type 2 | Result type |
| --- | --- | --- |
| AnsiChar | AnsiChar | AnsiChar |
| AnsiString | AnsiString, AnsiChar | AnsiString |
| Array | Array | Array of the same element type. |
| Boolean types | Boolean types | Common Boolean type or Boolean type. |
| Class reference | Class reference, nil | Common type, Base type, or TClass. |
| Dynamic array | Dynamic array | Common dynamic array type. |
| Enum type | Enum type | Common enum type. |
| File | File | Error (use pointer to File type). |
| Instance | Instance, nil | Common type, Base type, or TObject. |
| Integral types | Integral types | Common integral type, or the type that covers both ranges. |
| Integral types | Real types | Real type. |
| Interface | Interface, nil | Common type, Base type, or IInterface. |
| nil | AnsiChar, ShortString, AnsiString, WideChar, WideString | Pointer to WideChar |
| Object | Object | Object type. |
| Pointer to AnsiChar | AnsiChar, ShortString, AnsiString literal | Pointer to AnsiChar |
| Pointer to AnsiChar | Pointer to WideChar literal | Pointer to WideChar |
| Pointer to WideChar | AnsiChar, ShortString, AnsiString, WideChar, WideString, UnicodeString | UnicodeString |
| Pointer types | Pointer types | Common pointer type, or Pointer type. |
| Procedure | Procedure | Procedure type, if they have the same signature. |
| Procedure of object | Procedure of object | Procedure for object type if they have the same signature. |
| Real types | Real types | The real type with the greater precision. |
| Record | Record | Record type. |
| Reference to procedure Anonymous Method | Reference to procedure Anonymous Method | Reference to the procedure type, if they have the same signature. |
| Set | Set | Common set type, or large set type if list literals. |
| Short string | Short string, AnsiChar | Longer short string |
| Text | Text | Error (use pointer to Text type). |
| UnicodeString | UnicodeString, WideString, AnsiString, Short string, WideChar, AnsiChar | UnicodeString |
| Variant | Variant | Variant type. |
| WideChar | WideChar | WideChar |
| WideChar | AnsiChar, AnsiString | UnicodeString |
| WideString | WideString, AnsiString, WideChar, AnsiChar | WideString |

## Reality

Unfortunately, the documented operand/result type combinations are inaccurate or incomplete in various ways.
Below is what Delphi 13 actually does.

| Type 1 | Type 2 | Result type |
| --- | --- | --- |
| AnsiChar | AnsiChar | AnsiChar |
| AnsiString | AnsiString, AnsiChar | Left operand (code pages don't unify) |
| Array | Array | Identical type only; else error |
| Boolean types | Boolean types | Left operand, never widened |
| Class reference | Class reference, nil | Reference to nearest common ancestor; same class -> **right** operand; nil -> the reference |
| Dynamic array | Dynamic array, nil | Identical type only (nil -> the array); else error |
| Dynamic array | `[...]` constructor | Error |
| Enum type | Enum type | The enum (subranges unify to it); distinct enums error |
| File | File | Error (use pointer to File type). |
| Instance | Instance, nil | Nearest common ancestor class, never an interface; strong aliases are siblings of their base; nil -> the class |
| Instance | Interface | Error |
| Integral types | Integral types | Narrowest type covering both ranges; signed/unsigned tie -> anonymous unsigned subrange, or UInt64 |
| Integral types | Real types | Extended; Comp/Currency where Extended is 8 bytes |
| Interface | Interface, nil | Nearest common ancestor; nil -> the interface |
| nil | AnsiChar, ShortString, AnsiString, WideChar, WideString | Error |
| nil | Procedure, procedure of object, reference to | The procedural type |
| nil | `[...]` constructor | Anonymous `array of T` |
| nil | Variant, any value type | Error |
| Object | Object | Identical type only; else error |
| Pointer to AnsiChar | AnsiChar, ShortString, AnsiString literal | Error |
| Pointer to AnsiChar | Pointer to WideChar literal | Error |
| Pointer to WideChar | AnsiChar, ShortString, AnsiString, WideChar, WideString, UnicodeString | UnicodeString |
| Pointer types | Pointer types | Left operand if either is untyped or the right is nil; nil on the left, or unrelated pointers -> Pointer |
| Procedure | Procedure | Left operand, if they have the same signature |
| Procedure of object | Procedure of object | Left operand, if they have the same signature |
| Mixed procedural kinds | Mixed procedural kinds | Error |
| Real types | Real types | Extended; Comp/Currency where Extended is 8 bytes |
| Record | Record | Identical type only; else error |
| Reference to procedure Anonymous Method | Reference to procedure Anonymous Method | The `reference to` type, if they have the same signature |
| Set | Set | The containing set (left on ties); else `set of Byte` or `set of AnsiChar` |
| Set | `[...]` constructor | The set, if the elements are compatible |
| `[...]` constructor | `[...]` constructor | Left operand's set reading, if all elements are ordinal; else error |
| ShortString | ShortString, AnsiChar | Longer ShortString |
| Text | Text | Error (use pointer to Text type). |
| UnicodeString | UnicodeString, WideString, AnsiString, Short string, WideChar, AnsiChar | Left operand |
| Variant | Variant | Variant (even for OleVariant mixes) |
| WideChar | WideChar | WideChar |
| WideChar | AnsiChar, AnsiString | UnicodeString |
| WideString | WideString, AnsiString, WideChar, AnsiChar | Left operand |

Working out the actual rules required a lot of painful testing against the Delphi 13 compiler, but I think this is pretty accurate now.

It definitely isn't a "least upper bound" as commonly understood, or as documented:
  - operand types can end up stuffed into ill-fitting result types
    - signed integers going into `UInt64`
    - large booleans going into smaller booleans (whichever the left operand happened to be)
  - ties typically go to the left operand, so ordering of the expression is significant

## Fun Facts

- `if b then [] else DynArray` crashes the 64-bit compiler with an internal error (F2084).

### DIFF
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/visitors/CognitiveComplexityVisitor.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/visitors/CognitiveComplexityVisitor.java
@@ -31,6 +31,7 @@ import org.sonar.plugins.communitydelphi.api.ast.DelphiNode;
 import org.sonar.plugins.communitydelphi.api.ast.ExceptBlockNode;
 import org.sonar.plugins.communitydelphi.api.ast.ExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.ForStatementNode;
+import org.sonar.plugins.communitydelphi.api.ast.IfExpressionNode;
 import org.sonar.plugins.communitydelphi.api.ast.IfStatementNode;
 import org.sonar.plugins.communitydelphi.api.ast.RepeatStatementNode;
 import org.sonar.plugins.communitydelphi.api.ast.StatementNode;
@@ -85,6 +86,31 @@ public class CognitiveComplexityVisitor implements DelphiParserVisitor<Data> {
       if (bareElse) {
         --data.nesting;
       }
+    }
+
+    return data;
+  }
+
+  @Override
+  public Data visit(IfExpressionNode expression, Data data) {
+    data.increaseComplexityByNesting();
+    expression.getGuardExpression().accept(this, data);
+
+    ++data.nesting;
+    expression.getThenExpression().accept(this, data);
+    --data.nesting;
+
+    ExpressionNode elseBranch = expression.getElseExpression();
+    boolean bareElse = !(elseBranch.skipParentheses() instanceof IfExpressionNode);
+    if (bareElse) {
+      data.increaseComplexityByOne();
+      ++data.nesting;
+    } else {
+      data.updateComplexity(data.nesting - 1);
+    }
+    elseBranch.accept(this, data);
+    if (bareElse) {
+      --data.nesting;
     }
 
     return data;
--- a/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ArrayElementTypeResolver.java
+++ b/delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ArrayElementTypeResolver.java
@@ -1,0 +1,161 @@
+/*
+ * Sonar Delphi Plugin
+ * Copyright (C) 2026 Integrated Application Development
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package au.com.integradev.delphi.symbol.resolve;
+
+import static org.sonar.plugins.communitydelphi.api.type.TypeFactory.unknownType;
+
+import au.com.integradev.delphi.type.TypeUtils;
+import java.math.BigInteger;
+import java.util.List;
+import org.sonar.plugins.communitydelphi.api.type.IntrinsicType;
+import org.sonar.plugins.communitydelphi.api.type.Type;
+import org.sonar.plugins.communitydelphi.api.type.Type.ClassReferenceType;
+import org.sonar.plugins.communitydelphi.api.type.Type.IntegerType;
+import org.sonar.plugins.communitydelphi.api.type.Type.PointerType;
+import org.sonar.plugins.communitydelphi.api.type.TypeFactory;
+
+/** Resolves the element type the compiler infers for an array literal. */
+final class ArrayElementTypeResolver {
+  private final TypeFactory typeFactory;
+
+  ArrayElementTypeResolver(TypeFactory typeFactory) {
+    this.typeFactory = typeFactory;
+  }
+
+  Type elementType(List<Type> elements) {
+    return elements.stream().map(this::normalize).reduce(this::combine).orElse(unknownType());
+  }
+
+  private Type normalize(Type type) {
+    if (type.isReal()) {
+      return typeFactory.getIntrinsic(IntrinsicType.EXTENDED);
+    }
+    if (isNilPointer(type)) {
+      return typeFactory.getIntrinsic(IntrinsicType.POINTER);
+    }
+    return type;
+  }
+
+  private Type combine(Type first, Type second) {
+    if (first.is(second)) {
+      return first;
+    }
+
+    if (first.isInteger() && second.isInteger()) {
+      return combineInteger((IntegerType) first, (IntegerType) second);
+    }
+
+    if (isTextual(first) && isTextual(second)) {
+      return typeFactory.getIntrinsic(IntrinsicType.UNICODESTRING);
+    }
+
+    if (isBooleanBased(first) && isBooleanBased(second)) {
+      return typeFactory.getIntrinsic(IntrinsicType.BOOLEAN);
+    }
+
+    Type firstBase = TypeUtils.findBaseType(first);
+    if (firstBase.isEnum() && firstBase.is(TypeUtils.findBaseType(second))) {
+      return firstBase;
+    }
+
+    if ((first.isClass() && second.isClass()) || (first.isInterface() && second.isInterface())) {
+      if (second.isDescendantOf(first)) {
+        return first;
+      }
+      if (first.isDescendantOf(second)) {
+        return second;
+      }
+      return rootOf(first);
+    }
+
+    if (first.isClassReference() && second.isClassReference()) {
+      return combineClassReference(first, second);
+    }
+
+    if (first.isPointer() && second.isPointer()) {
+      return typeFactory.getIntrinsic(IntrinsicType.POINTER);
+    }
+
+    return unknownType();
+  }
+
+  private Type combineInteger(IntegerType first, IntegerType second) {
+    IntegerType left = promote(first);
+    IntegerType right = promote(second);
+    if (left.is(right)) {
+      return left;
+    }
+
+    BigInteger min = left.min().min(right.min());
+    BigInteger max = left.max().max(right.max());
+    for (IntrinsicType candidate :
+        List.of(
+            IntrinsicType.INTEGER,
+            IntrinsicType.CARDINAL,
+            IntrinsicType.INT64,
+            IntrinsicType.UINT64)) {
+      IntegerType type = (IntegerType) typeFactory.getIntrinsic(candidate);
+      if (type.min().compareTo(min) <= 0 && type.max().compareTo(max) >= 0) {
+        return type;
+      }
+    }
+
+    return typeFactory.getIntrinsic(IntrinsicType.UINT64);
+  }
+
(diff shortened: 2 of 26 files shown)


delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/visitors/CognitiveComplexityVisitor.java
```diff
@@ -90,6 +91,19 @@ public Data visit(IfStatementNode statement, Data data) {
...
   }
 
+  @Override
+  public Data visit(IfExpressionNode expression, Data data) {
+    data.increaseComplexityByNesting();
+    expression.getGuardExpression().accept(this, data);
+
+    ++data.nesting;
+    expression.getThenExpression().accept(this, data);
+    expression.getElseExpression().accept(this, data);
```
fourls: For if statements, we have special handling for `else` blocks directly containing an `if` block. This is to prevent over-counting cognitive complexity for `else if` statements. For example:

```delphi
if Foo then
  // ...
else // complexity+0 (not bare)
  if Bar then
    // ...
  else // complexity+0 (not bare)
    if Baz then
    else // complexity+1 (bare)
      // ...
```

We should do the same `else` handling for if expressions, e.g.

```delphi
//                      +0                 +0                 +1            
Result := if Foo then A else if Bar then B else if Baz then C else D;
```

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolver.java
```diff
@@ -0,0 +1,476 @@
...
+    if (first.isSet() && second.isSet()) {
+      return commonSetType(first, second);
+    }
+
+    if (first.isSet() && second.isArrayConstructor()) {
+      return setAbsorbingConstructor((CollectionType) first, (ArrayConstructorType) second);
+    }
+    if (second.isSet() && first.isArrayConstructor()) {
+      return setAbsorbingConstructor((CollectionType) second, (ArrayConstructorType) first);
+    }
```
fourls, line 146: Redundant cast:

```suggestion
    if (first.isSet() && second.isArrayConstructor()) {
      return setAbsorbingConstructor(first, (ArrayConstructorType) second);
    }
    if (second.isSet() && first.isArrayConstructor()) {
      return setAbsorbingConstructor(second, (ArrayConstructorType) first);
    }
```

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolverTest.java
```diff
@@ -0,0 +1,1051 @@
...
+    Type enumA = factory.enumeration("TEnumA", DelphiScope.unknownScope());
+    Type enumB = factory.enumeration("TEnumB", DelphiScope.unknownScope());
+    Type setA = FACTORY.set(enumA);
+
+    assertThat(resolve(setA, FACTORY.set(enumA))).isEqualTo(setA);
+    assertThat(resolve(setA, FACTORY.set(enumB))).matches(Type::isUnknown);
+  }
+
+  @Test
+  void testProceduralBranchesWithDifferentSignaturesResolveToUnknown() {
```
fourls, line 644: To improve coverage, let's add

- procedural branches with different function return types resolve to unknown
- branches with procedure and function resolves to unknown
- procedural branches with the same number of arguments but different types resolves to unknown

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolverTest.java
```diff
@@ -0,0 +1,1051 @@
...
+    Type dynamicArray =
+        factory.array(null, intrinsic(IntrinsicType.BYTE), Set.of(ArrayOption.DYNAMIC));
+    Type set = FACTORY.set(intrinsic(IntrinsicType.BYTE));
+
+    assertThat(resolve(dynamicArray, set)).matches(Type::isUnknown);
+    assertThat(resolve(set, dynamicArray)).matches(Type::isUnknown);
+  }
+
+  @Test
+  void testOrdinalSetConstructorsResolveToTheFirstConstructor() {
```
fourls: To improve coverage, add cases for other ordinal types (integer, enum, char, boolean, subrange).

fourls (changes requested):
Looks good for the most part! A couple of minor comments.

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolverTest.java
cirras: Agreed, will do.

delphi-frontend/src/test/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolverTest.java
cirras: Agreed, will do.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolver.java
cirras: Good catch, this was leftover from a refactor.

delphi-frontend/src/main/java/au/com/integradev/delphi/antlr/ast/visitors/CognitiveComplexityVisitor.java
cirras: I was originally up in the air on this one and probably should have mentioned it. I guess I thought nesting if-expressions was just too ✨crazy✨ and nobody would actually do it.

... but on reflection, of course they will. And when they do, it won't make the code *that* much more complex (even if I think it's silly and ugly.)

So yeah, I'll go ahead and do the same special handling for `if` expressions. Later, it might be worth adding a rule to forbid nesting them (N) deep.

cirras:
Addressed review feedback and, in the process, discovered the resolver was handling array constructors incorrectly.
Pulling that thread a little further, it turns out the `TypeInferrer` was also handling them incorrectly. That was a fun incidental discovery. 😩
Patched up in a separate commit.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolver.java
```diff
@@ -0,0 +1,477 @@
...
+
+    if (ancestor.is(firstReference.classType())) {
+      return firstReference;
+    }
+
+    if (ancestor.is(secondReference.classType())) {
+      return secondReference;
+    }
+
+    return ((TypeFactoryImpl) typeFactory).classOf(null, ancestor);
```
fourls: Sorry to report another redundant cast.

```suggestion
    return typeFactory.classOf(null, ancestor);
```

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ArrayElementTypeResolver.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.List;
+import org.sonar.plugins.communitydelphi.api.type.IntrinsicType;
+import org.sonar.plugins.communitydelphi.api.type.Type;
+import org.sonar.plugins.communitydelphi.api.type.Type.ClassReferenceType;
+import org.sonar.plugins.communitydelphi.api.type.Type.IntegerType;
+import org.sonar.plugins.communitydelphi.api.type.Type.PointerType;
+import org.sonar.plugins.communitydelphi.api.type.TypeFactory;
+
+/** Resolves the element type the compiler infers for an array literal. */
+final class ArrayElementTypeResolver {
```
fourls: Maybe `ArrayElementInferredTypeResolver`? The name makes it sound a bit more general than it is.

fourls (changes requested):
Looks really good! The array element type inference being wrong is a bit of a surprise - I guess it's a fairly niche case. 

On top of the comments below, could you also rebase on master to get the Java 25 compatibility fixes?

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/CommonTypeResolver.java
cirras: Oops - fixed.

delphi-frontend/src/main/java/au/com/integradev/delphi/symbol/resolve/ArrayElementTypeResolver.java
cirras: I hate my name, I hate your name, I hate everyone's name.
Let's compromise and pick yours.

fourls (approved):
Looks good! This will go a long way to unblocking Delphi 13 support. 

As always, I'm sure there's a bunch of bizarre edge case behavioural quirks that we've missed and will have to slowly discover over time 😅

PR #451, opened 2026-08-04
Migrate to JDK 25 and upgrade various dependencies

cirras:
This PR migrates the project from JDK 17 to JDK 25, upgrading most dependencies in the process.

Upgrades:
- Required JDK (17 -> 25)
- Required Maven (3.3.9 -> 3.6.3)
- Various maven libraries
- Various maven plugins

PR #453, opened 2026-08-07
Fix format workflow syntax

fourls:
Fixes a syntax error in the format action's workflow file, inadvertently introduced in #451.


cirras (approved):
No idea how I managed to delete this line by mistake. 😅

PR #454, opened 2026-08-10
Implement `is not` and `not in` operators

### LINKED ISSUE #408: Support `is not` operator
### Prerequisites

### Engine area

Delphi language support

### Improvement description

Delphi 13 has introduced a new `is not` operator.

## Example
```pas
if not (Obj is TFoo) then ... // classic syntax
if Obj is not TFoo then ... // new syntax
```

## See
- [The is not and not in Operators](https://docwiki.embarcadero.com/RADStudio/Florence/en/Expressions_(Delphi)#The_is_not_and_not_in_Operators)

### Rationale

Delphi 13 support.

### LINKED ISSUE #409: Support `not in` operator
### Prerequisites

### Engine area

Delphi language support

### Improvement description

Delphi 13 has introduced a new `not in` operator.

This might require some extra care to ensure operator precedence is respected properly.

## Example
```pas
if not (x in [Apple, Orange] then ... // classic syntax
if x not in [Apple, Orange] then ... // new syntax
```

## See
- [The is not and not in Operators](https://docwiki.embarcadero.com/RADStudio/Florence/en/Expressions_(Delphi)#The_is_not_and_not_in_Operators)
- [Operator Precedence](https://docwiki.embarcadero.com/RADStudio/Florence/en/Expressions_(Delphi)#Operator_Precedence)

### Rationale

Delphi 13 support.

cirras:
This PR implements the new binary operators added in Delphi 13.

In terms of parsing, operator precedence, and type resolution, this wasn't a difficult implementation. There were other challenges.

"Compound" operators with multiple tokens make previous AST modeling of binary and unary expression nodes a little tricky. The operators in these expressions are now modeled as their own typed child nodes.

This required me to build binary expressions a little differently, which made me drop the `resetBinaryExpressionTokens` fix introduced in 041a0d9. In doing so, I had to wrestle with node text range calculation for (probably) the 27th time. The new shape seems to work well.

Closes #408, #409.


delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -362,6 +366,17 @@ import au.com.integradev.delphi.utils.LocatableException;
     result.setStartIndex(firstToken.getStartIndex());
     result.setLine(firstToken.getLine());
     result.setCharPositionInLine(firstToken.getCharPositionInLine());
+    if (firstToken.getInputStream() == lastToken.getInputStream()) {
+      // A null text makes CommonToken#getText() recompute the image from the start/stop indexes.
+      result.setText(null);
+    } else {
+      // The tokens straddle an include file boundary, so there is no single source span to read.
+      StringBuilder text = new StringBuilder();
```
fourls: Since we know the exact size of the string, let's set the `StringBuilder` capacity here:

```suggestion
      StringBuilder text = new StringBuilder(count);
```

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
```diff
@@ -924,18 +924,23 @@ expression                   : relationalExpression
                              | anonymousMethod
                              | ifExpression
                              ;
-// ANTLR sets the begin and end tokens for nested binary expression nodes
-// in relationalOperator, not relationalExpression, meaning that their
-// token range only contains the operator. resetBinaryExpressionTokens is needed
-// to reset the start and end tokens so that they must be recalculated
-// when retrieved (i.e., after their children have been correctly assigned).
-relationalExpression         : additiveExpression (relationalOperator^ additiveExpression)* { resetBinaryExpressionTokens(root_0); }
+relationalExpression         : (lhs=additiveExpression -> $lhs)
```
fourls, line 927: What's the purpose of the `lhs=... -> $lhs` mapping here and in the other rules? Why not just e.g. `additiveExpression` like previously?

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/directive/expression/ExpressionParser.java
fourls, line 1: All these changes to support `NOT IN` in the ExpressionParser look good, but `IS` and `IS NOT` also need to be supported - they are valid in compiler directives, though seem to blindly return true. The following compiles and runs, and prints both lines. 

```delphi
program foo;

var
  Obj: TObject;
begin
  {$if Obj is TObject}
  WriteLn('is');
  {$endif}

  {$if Obj is not TObject}
  WriteLn('is not');
  {$endif}

  ReadLn;
end.
```

CHANGELOG.md
```diff
@@ -9,14 +9,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - **API:** `IfExpressionNode` node type.
+- **API:** `OperatorNode` node type.
+- **API:** `BinaryOperatorNode` node type.
+- **API:** `UnaryOperatorNode` node type.
+- **API:** `DelphiTokenType.BINARY_EXPRESSION` token type.
+- **API:** `DelphiTokenType.UNARY_EXPRESSION` token type.
+- **API:** `DelphiTokenType.IS_NOT` token type.
+- **API:** `DelphiTokenType.NOT_IN` token type.
+- **API:** `BinaryOperator.IS_NOT` enum value.
+- **API:** `BinaryOperator.NOT_IN` enum value.
```
fourls, line 22: Changelog should also include the addition of `BinaryExpressionNode::getOperatorNode()`.

delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
```diff
@@ -34,7 +33,7 @@
 public class ExplicitBitwiseNotCheck extends DelphiCheck {
   @Override
   public DelphiCheckContext visit(BinaryExpressionNode node, DelphiCheckContext context) {
-    if (node.getOperator() == BinaryOperator.IN) {
+    if (node.getOperator() == BinaryOperator.IN || node.getOperator() == BinaryOperator.NOT_IN) {
```
fourls: I don't know if this is a necessary change... This rule is making up for `if not A in B then` looking correct to people without a strong knowledge of Delphi precedence. I don't think `if not A not in B then` really has the same problem - to me that actually does read correctly as a bitwise not followed by a logical not. It's sort of a different class of mistake and including it in this rule muddies the waters a little bit.

fourls (changes requested):

delphi-checks/src/main/java/au/com/integradev/delphi/checks/ExplicitBitwiseNotCheck.java
cirras: Reasonable take, I agree and will back this one out.

CHANGELOG.md
cirras: Good catch.

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/directive/expression/ExpressionParser.java
cirras: I didn't expect this at all since it makes no real sense at compile-time, which is why `is` isn't supported to begin with.
I'll take a closer look.

fourls: Yeah, it's an odd one...

delphi-frontend/src/main/antlr3/au/com/integradev/delphi/antlr/Delphi.g
cirras: Ugh, I think there was an arcane antlr3 codegen reason... But that might have only applied to my previous iteration of this annoying expression construction refactor.
I'll give it a second look and get back to you.

cirras: Okay, so basically once a grammar rule alternative contains any `->` rewrite, antlr3 stops auto-building the tree and diverts everything it matches into rewrite streams instead.

`^(TkBinaryExpression $relationalExpression $op $rhs)` gets built out of "the rule's result so far" and that initial `$lhs` rewrite is what puts the left operand into that result. Without it, you leave the left operand behind in an unread stream and the result tree is never built at all. iirc that leads to a mysterious `RewriteEmptyStreamException`.

fourls: Bizarre... that does ring a bell now that you mention it. All good!

delphi-frontend/src/main/java/au/com/integradev/delphi/preprocessor/directive/expression/ExpressionParser.java
cirras: This was a very neat little nightmare that forced me to split unevaluable expressions into 2 categories based on a mix of compiler behavior probing and respecting the SonarDelphi canon (fanfiction preprocessor implementation).

cirras: Oh, and I also randomly found out that we weren't handling string ordering comparisons, so that's been fixed in a separate commit as well.

fourls (approved):
Looks good!

PR #457, opened 2026-08-31
Fix obscure symbol reassociation bug

cirras:
This PR fixes an obscure bug related to the 2-pass semantic analysis we're doing where we build the symbol table in one pass, throw away ASTs as we go, and re-parse source files (reassociating symbol information) in a second pass.

For symbol reassociation purposes, we were registering declarations, occurences, and scopes against token indices as if they were unique at the node level. They're not.

This led to some tricky situations where scopes in particular would be incorrectly attributed to the wrong nodes. This bug was masked very well because most name resolution occurs during symbol table construction.

The problem came about after symbol reassociation. In the analysis phase, on-demand name resolution occurs during expression type resolution. Any analysis rule calling `ExpressionNode::getType` was potentially impacted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Incorrect token image on `>=` and `<=` operators.
 - Incorrect branch selection around `{$IF}` conditions that can't be evaluated at compile time.
+- Obscure bug where name resolution could occur in the wrong lexical scope.
 
 ## [1.20.0] - 2026-08-12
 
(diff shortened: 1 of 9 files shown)


cirras:
Fixed the failing tests and added a new pile of unit tests covering the `SymbolAssociationVisitor`.

Tests were failing because we were actually *relying* on this awful "multiple nodes return the same token index behavior" in several ways. Most notably, the registered location for every name occurrence is an `IdentifierNode`, but we were reattaching via the containing `NameReferenceNode` which had the same token index.

fourls (dismissed):
I think this looks good! 
As far as I can see we don't rely on this shared-index behaviour anywhere else. 
The new `SymbolAssociationVisitor` tests are really good to have.

cirras:
Realized I did the classic "wrong year in license header" mistake with the new file.
@fourls, can you reapprove?

fourls (approved):
I didn't notice, good catch 😆

PR #467, opened 2026-09-21
Fix parsing errors around attributes on visibility sections

### LINKED ISSUE #461: Attributes on visibility sections cause parsing failures
### Prerequisites

- [x] This bug is in SonarDelphi, not SonarQube or my Delphi code.
- [x] This bug has not already been reported.

### SonarDelphi version

1.20

### Delphi version

13

### SonarQube version

_No response_

### Issue description

The parsers only supports the attribute before the field name like this:
`  private [MyAttribute] FMyField: String;
    
  private var [MyAttribute] FMyField: String;`

but RAD supports those too:

`  [MyAttribute]
  private FMyField: String;

  [MyAttribute]	
  private var FMyField: String;`

### Steps to reproduce

Create a class with attribute before visibility

`  [MyAttribute]
  private FMyField: String;

  [MyAttribute]	
  private var FMyField: String;`

sonar-delphi is unable to parse the class.

_No response_

Bazooper-blip:
This PR fixes parsing errors on attributes that precede visibility specifiers or `var` sections in class bodies.

These attribute lists are parsed as standalone items and their names are resolved, but they are not yet associated with the next declared member (per @fourls's findings in the issue). I've left that out of scope here.

I don't have access to a Delphi 13 compiler at the moment, so the test fixture is limited to the forms confirmed in the issue thread.

Fixes #461.


cirras (approved):
Looks good. Thanks!

